PR #50, opened 2026-03-28
fix: connect-existing with remote/containerized Firefox

MayCXC:
## Summary

Fixes several issues that prevented `--connect-existing` from working with Firefox running on a remote host or in a separate container (e.g., via SSH tunnel).

### Bug fixes

1. **`--marionette-host` CLI arg was silently ignored.** The option was parsed and the type had the field, but it was never passed to the options object. Geckodriver always connected to `127.0.0.1`.

2. **selenium-manager downloaded a 307MB Firefox binary.** In connect-existing mode Firefox is already running. Changed `--browser firefox` to `--driver geckodriver` so selenium-manager only downloads the driver (~5MB).

3. **No cleanup on disconnect.** No lifecycle handling existed. The Marionette session was never released when the MCP client disconnected, so Firefox refused new connections until restarted. Added SIGTERM/SIGINT/stdin handlers that clean up Firefox and close the server before exiting, following the pattern from the [MCP SDK examples](https://github.com/modelcontextprotocol/typescript-sdk/tree/main/examples/server/src). `StdioServerTransport` does not currently fire `onclose` on stdin EOF ([PR #1814](https://github.com/modelcontextprotocol/typescript-sdk/pull/1814)), so raw stdin listeners are included as a workaround.

4. **`kill()` skipped session cleanup.** Was synchronous and did not send `DELETE /session`. Made async with session cleanup so Marionette accepts new connections.

5. **Seamless reconnect after Firefox restart.** `getFirefox()` now detects a lost connection and reconnects transparently instead of throwing an error on the first call.

### New features

6. **`--marionette-host` parameter.** Controls which host geckodriver connects to for Marionette and where the BiDi WebSocket URL is rewritten to point. Supports `MARIONETTE_HOST` env var.

7. **BiDi support for connect-existing.** Opens a WebSocket to Firefox's Remote Agent using the `webSocketUrl` session capability, enabling console and network event monitoring.

## Test plan

- [x] `list_pages` works with `--connect-existing --marionette-host host.internal`
- [x] `list_console_messages` returns BiDi data
- [x] Quit and restart Firefox without MCP reconnect: seamless reconnect on next tool call, pages and BiDi both work
- [x] MCP reconnect without quitting Firefox: old container exits cleanly (no zombie), new container connects, pages and BiDi both work
- [x] No stale Marionette sessions after disconnect

Supersedes #51.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -110,6 +110,11 @@ export const cliOptions = {
     description: 'Marionette port to connect to when using --connect-existing (default: 2828)',
     default: Number(process.env.MARIONETTE_PORT ?? '2828'),
   },
+  marionetteHost: {
+    type: 'string',
+    description: 'Marionette host to connect to when using --connect-existing (default: 127.0.0.1). Also used as the BiDi WebSocket connect address when different from 127.0.0.1.',
+    default: process.env.MARIONETTE_HOST ?? '127.0.0.1',
+  },
   env: {
     type: 'array',
     description:
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -8,6 +8,7 @@ import { spawn, type ChildProcess } from 'node:child_process';
 import { mkdirSync, openSync, closeSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { join } from 'node:path';
+import WebSocket from 'ws';
 import type { FirefoxLaunchOptions } from './types.js';
 import { log, logDebug } from '../utils/logger.js';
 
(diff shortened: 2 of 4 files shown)


juliandescottes:
@MayCXC 

Hi Aaron,

Already sent an email, but trying here as well. We want to update the project's license to Apache/MIT https://bugzilla.mozilla.org/show_bug.cgi?id=2032966

We need the agreement from existing contributors to proceed. If you are OK with the change, could either reply here or by email to say that you agree. Let me know if you have any question.

MayCXC:
> @MayCXC
> 
> Hi Aaron,
> 
> Already sent an email, but trying here as well. We want to update the project's license to Apache/MIT https://bugzilla.mozilla.org/show_bug.cgi?id=2032966
> 
> We need the agreement from existing contributors to proceed. If you are OK with the change, could either reply here or by email to say that you agree. Let me know if you have any question.

OK with me 👍

juliandescottes:
Thanks a lot!

PR #65, opened 2026-04-03
refactor: rewrite connect-existing to rely on selenium --connect-existing feature

juliandescottes:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=2029130

With this PR the connect-existing mode reuses the same Driver class as if Firefox is started by the MCP.
We can remove `GeckodriverElement` and `GeckodriverHttpDriver`. 

If the browser was started without --remote-debugging-port (eg. only with --marionette) only WebDriver classic capabilities will be available. If started with both --remote-debugging-port and --marionette, all WebDriver BiDi features should also be available.


juliandescottes:
Alright should be ready for final review @padenot :
- removed the custom host support
- also removed leftover interfaces which were no longer required since we directly get WebDriver instances from Selenium in all situations.

padenot (approved):

PR #68, opened 2026-04-17
feature: --profile-path should use a subfolder and warn about reusing existing profiles

juliandescottes:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=2032351

This PR makes --profile-path safer by routing Firefox through a dedicated firefox_devtools_mcp_profile/ subfolder instead of using the provided path directly. This prevents MCP from accessing a real Firefox profile (with passwords, cookies, bookmarks) in an automated session.

On first use, the subfolder is created and prefs.js is copied from the parent so user preferences carry over. 

If the provided path looks like a real Firefox profile, a warning is surfaced to the user via the MCP tool response.

### DIFF
--- a/src/firefox/profile.ts
+++ b/src/firefox/profile.ts
@@ -1,0 +1,70 @@
+/**
+ * Profile path resolution for Firefox DevTools MCP
+ *
+ * Rather than using a user-provided path directly as the Firefox profile,
+ * we create a dedicated subfolder. This prevents accidentally reusing a real
+ * browser profile (with bookmarks, saved passwords, cookies, …) inside an
+ * automated/remote-accessible session.
+ */
+
+import { existsSync, mkdirSync, copyFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { log } from '../utils/logger.js';
+
+export const MCP_PROFILE_DIR_NAME = 'firefox_devtools_mcp_profile';
+
+/**
+ * Files that are characteristic of an existing Firefox profile directory.
+ * Their presence indicates the user may have pointed at their real profile.
+ */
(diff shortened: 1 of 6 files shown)


juliandescottes:
Flagged everyone for review. This is a slight regression in terms of reusing existing profiles, and I'm not against having an escape hatch if we consider that reusing profiles as-is is a must. But IMO it's a huge liability and we should be careful with that.

src/firefox/profile.ts
```diff
@@ -0,0 +1,77 @@
...
+import { join } from 'node:path';
+import { log } from '../utils/logger.js';
+
+export const MCP_PROFILE_DIR_NAME = 'firefox_devtools_mcp_profile';
+
+/**
+ * Files that are characteristic of an existing Firefox profile directory.
+ * Their presence indicates the user may have pointed at their real profile.
+ */
+const FIREFOX_PROFILE_INDICATORS = ['prefs.js', 'places.sqlite', 'cert9.db', 'key4.db'];
```
padenot, line 20: I'm thinking we could do it a slightly different way: allowing to create a profile (like we do here), but drop a flag in there (e.g. simply a file `.firefox-devtools-mcp`) that would mark this profile as being used by this MCP. This profile will be able to be reused without problem.

I think in development it's useful to keep some state sometimes. We can think about this more and do something to support that use case as a follow-up.

padenot (approved):
Maybe a follow-up, but this will work for now and is better than the current state.

freema (approved):

src/firefox/profile.ts
juliandescottes: I agree, that would be better than checking random files. I will file a follow up bug.

juliandescottes: https://bugzilla.mozilla.org/show_bug.cgi?id=2034077

PR #70, opened 2026-04-20
fix(claude): correct network status type mismatch in debug command

pngdeity:
The 'debug' command in the Claude Code plugin instructed the model to use `status="failed"` for network requests, which conflicted with the underlying tool's numerical schema in `src/tools/network.ts`.

This PR updates the command definition to use `statusMin=400`, which correctly captures all 4xx and 5xx errors and aligns with the existing implementation.

Verified using a reproduction script against the tool handler.

### DIFF
--- a/plugins/claude/firefox-devtools/commands/debug.md
+++ b/plugins/claude/firefox-devtools/commands/debug.md
@@ -26,7 +26,7 @@ Displays debugging information from the current page.
 ## What Happens
 
 - `console`: Calls `list_console_messages` with `level="error"`
-- `network`: Calls `list_network_requests` with `status="failed"`
+- `network`: Calls `list_network_requests` with `statusMin=400`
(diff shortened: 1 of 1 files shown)


juliandescottes:
@pngdeity Thanks for the PR, that looks correct to me. If you want to move it from draft to review, happy to merge!

juliandescottes:
@pngdeity thanks for the PR!

At the moment we are in the process of re-licensing the repository under the Apache/MIT dual license (as opposed to MIT only at the moment).

More details at https://bugzilla.mozilla.org/show_bug.cgi?id=2032966

Can you reply to this comment to confirm you agree with the upcoming license update?

pngdeity:
Hi @juliandescottes, I can confirm that I agree with the upcoming license update from strictly MIT to a dual-licensed Apache/MIT.

juliandescottes (approved):
Thanks for the fix!

PR #72, opened 2026-04-22
chore: move test scripts to scripts/

juliandescottes:
This PR moves 4 `test-*.mjs` files from the root folder of the repo to the scripts subfolder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,15 @@
     "test:watch": "vitest watch",
     "test:ui": "vitest --ui",
     "test:tools": "node scripts/test-bidi-devtools.js",
-    "test:input": "node scripts/test-input-tools.js",
-    "test:screenshot": "node scripts/test-screenshot.js",
+    "test:console": "node scripts/test-console.js",
     "test:dialog": "node scripts/test-dialog.js",
+    "test:firefox-temp": "node scripts/test-firefox-temp.js",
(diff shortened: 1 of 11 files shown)


dmose (approved):
r=dmose, with a couple of nits, which could easily be addressed in another commit/PR/bug if you'd rather not do it here:

* it would be nice to not require ${HOME}/firefox/firefox as a path.
* the way my account is configured at least, I need to explicitly invoke `node test-something.mjs` for the .mjs tests, but not for the .js tests. Why the two different naming schemes?

juliandescottes:
> r=dmose, with a couple of nits, which could easily be addressed in another commit/PR/bug if you'd rather not do it here:
> 
>     * it would be nice to not require ${HOME}/firefox/firefox as a path.
> 
>     * the way my account is configured at least, I need to explicitly invoke `node test-something.mjs` for the .mjs tests, but not for the .js tests. Why the two different naming schemes?

Good points, I mostly moved the files to cleanup the root, but we can cleanup a bit further.
- rename all from mjs to js
- add missing scripts to package.json (not touching TaskFile for now, that's not something I personally use but it seems like it leads to double maintenance?)
- switch to process.env.FIREFOX_PATH instead of relying on a random `${HOME}/firefox/firefox` 
- on that note, switch all manual tests to use process.env.FIREFOX_PATH consistently

PR #73, opened 2026-04-23
fix: use node to launch MCP server for Windows compatibility

MReschenberg:
On Windows, `npx` is a .cmd batch file that cannot be spawned directly by Claude Code's MCP server launcher (which uses child_process.spawn without a shell). Claude Code detects this and shows the warning:

  Windows requires 'cmd /c' wrapper to execute npx

Without a fix, the MCP server silently fails to connect -- tools appear registered but every call fails indefinitely.

Replace `npx -y firefox-devtools-mcp@latest` with `node` pointing directly at the installed `dist/index.js` via `${CLAUDE_PLUGIN_ROOT}`. If I'm understanding things correctly, I think `node` (node.exe) is a real executable on all platforms and should always be spawn-able directly without a shell wrapper. It'd be great if someone can verify this on a non-win machine though. 

I verified on Windows 11 with Claude Code CLI in moz-build shell, and separately in git bash.

### DIFF
--- a/plugins/claude/firefox-devtools/.mcp.json
+++ b/plugins/claude/firefox-devtools/.mcp.json
@@ -1,8 +1,8 @@
 {
   "mcpServers": {
     "firefox-devtools": {
-      "command": "npx",
-      "args": ["-y", "firefox-devtools-mcp@latest"]
+      "command": "node",
+      "args": ["${CLAUDE_PLUGIN_ROOT}/../../../dist/index.js"]
     }
   }
 }


juliandescottes:
I am not sure how to test this.
I think claude plugins need to be published on a marketplace first in order to be installed, but I can't find any marketplace where this was published? 

Did you manage to use `claude plugin install firefox-devtools`?

Otherwise the change seems sensible, but I will need to first understand if we publish this plugin anywhere, and if not, why we are maintaining / keeping this here.

juliandescottes:
Based on https://github.com/mozilla/firefox-devtools-mcp/commit/6983d8b8bdad1fd7b2116fd35cf5f4051fa5725b and subsequent commits around the same files, I feel like the plugin was never published anywhere? 

@freema would you know more about this?

juliandescottes (changes requested):
Moving to request changes, need feedback for my questions before reviewing.

freema:
@juliandescottes No, the plugin itself was never published as a Claude Code plugin not on the marketplace, not as a plugin package. The npm package is published (and works fine via npx), but the Claude Code plugin scaffold I added in https://github.com/mozilla/firefox-devtools-mcp/commit/6983d8b8bdad1fd7b2116fd35cf5f4051fa5725b was never finished. Up to you what to do delete it, finish publishing it as a plugin, or keep it for local dev only.

About this PR: the ${CLAUDE_PLUGIN_ROOT}/../../../dist/index.js path only works from a source checkout. A better fix for the Windows npx issue would be one of:

wrap with cmd /c on Windows ("command": "cmd", "args": ["/c", "npx", "-y", "..."]),
platform-specific config,
or actually publish it as a proper Claude Code plugin and let Claude Code handle the spawn properly.

juliandescottes:
@freema  Thanks for the info! It's fine to keep this around for a bit and see if we can publish it.

@MReschenberg I imagine you are using this from a firefox checkout. I don't know if you had issues with the default config but we are trying to improve it a bit in https://phabricator.services.mozilla.com/D294087

We could have a platform specific config documented here, but not sure what's the best way to do that.

MReschenberg:
> @MReschenberg I imagine you are using this from a firefox checkout. I don't know if you had issues with the default config but we are trying to improve it a bit in https://phabricator.services.mozilla.com/D294087

Oh cool, yes I am using this from a fx checkout. I can wait until that patch lands and revisit this if needed :) I'll close this for now

PR #76, opened 2026-04-27
chore: reduce emoji usage in MCP logs

juliandescottes:


juliandescottes:
I'm going to merge this, I think this does not require a full review.

PR #79, opened 2026-04-28
chore: update security document based on review feedback

juliandescottes:
Expand a bit the SECURITY documentation based on feedback from @ashmoz

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -24,7 +24,9 @@ Enables the `evaluate_script` tool, which lets the agent execute arbitrary JavaS
 
 ### `--enable-privileged-context`
 
-Enables tools that operate in Firefox's privileged (chrome) context: listing and selecting privileged contexts, evaluating privileged scripts, reading and writing Firefox preferences, and listing extensions. These tools require the `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1` environment variable to be set.
+Enables tools that operate in Firefox's privileged (chrome) context: listing and selecting privileged contexts, evaluating privileged scripts, reading and writing Firefox preferences, and listing extensions. These tools require the `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1` environment variable to be set, which is checked by the WebDriver implementation in Firefox to allow using any command that targets privileged contexts.
+
+Unless you are developing or modifying Firefox itself, you likely do not need this flag. To set preferences, you can always use the `--set-pref` command-line argument. If you are missing commands or features to debug web content, please file a bug on [Bugzilla](https://bugzilla.mozilla.org/enter_bug.cgi?format=__default__&blocked=2026717&product=Developer%20Infrastructure&component=Firefox%20MCP) or reach out in the [#firefox-devtools-mcp Matrix room](https://chat.mozilla.org/#/room/#firefox-devtools-mcp:mozilla.org).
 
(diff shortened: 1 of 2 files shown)


juliandescottes:
Added https://github.com/mozilla/firefox-devtools-mcp/pull/79/changes/08fb8dd04db2d552dd25eaf2c52b4a134bbf7abd to mention specifically the limitations of the claude sandbox. More should be added in https://bugzilla.mozilla.org/show_bug.cgi?id=2035840 but we first need to test it in more details.

SECURITY.md
```diff
@@ -4,22 +4,42 @@
...
 
-Avoid enabling `--enable-script` unless strictly necessary. The `evaluate_script` tool lets the agent execute arbitrary JavaScript in the page context, which significantly expands its attack surface.
+### `--enable-privileged-context`
 
-Keeping capabilities to the minimum needed reduces the potential impact of a compromised session or a prompt injection attack.
+Enables tools that operate in Firefox's privileged (chrome) context: listing and selecting privileged contexts, evaluating privileged scripts, reading and writing Firefox preferences, and listing extensions. These tools require the `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1` environment variable to be set.
+
+> **Warning:** When `--enable-privileged-context` is used together with `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1`, the agent gains access to privileged Firefox APIs with no web-content sandbox boundary. Depending on what the agent does with that access, this can extend to operating-system–level actions. Only use this combination in fully isolated environments.
```
dmose, line 31: Even though MOZ_REMOTE_ALLOW_SYSTEM_ACCESS isn't implemented by the MCP, it would be super helpful to link to (or include)  documentation for it here, so that it's easier to reason about when you would even need it at all, if ever.  

And maybe a follow-on ticket for mitigations for cases where it might be perceived to be needed.  This could be either documenting ways to avoid using, or implementation changes to make it unnecessary.

SECURITY.md
```diff
@@ -4,22 +4,42 @@
...
+
+### `--accept-insecure-certs`
+
+Disables TLS certificate validation, allowing the agent to visit sites with self-signed or expired certificates without warning. This removes a layer of authentication that would otherwise help detect man-in-the-middle scenarios.
+
+## Profile and Environment
+
+**Use a dedicated profile.** Never point the MCP server at your regular Firefox profile. Create a clean, separate profile for automation. This limits the data the agent can access and prevents a compromised session from touching your personal browsing data.
+
+**Consider a sandboxed environment.** For automation that involves untrusted content, or when `--enable-privileged-context` is required, run Firefox inside an isolated environment (a container, VM, or dedicated OS user account). This limits what an attacker can reach even if the agent is fully compromised.
```
dmose: Maybe even include something along the lines of "ideally with a proxy to enforce networking restrictions."

dmose (commented):
Here are some thoughts; I think it would be good to have @padenot also review this before it lands, if he's got bandwidth.

SECURITY.md
juliandescottes: AFAIK we don't have any documentation about this environment variable at the moment. I would slightly prefer having it in tree, but in the meantime I can add a short description here.

juliandescottes: Tried to add a light description of the variable + more details about the impacts + mitigations.

SECURITY.md
juliandescottes: Done

SECURITY.md
dmose: Looks good to me.

juliandescottes:
Thanks for the review Dan, I'm going to merge, I feel like Paul is very busy these days :)

PR #80, opened 2026-04-30
chore: add AGENTS.md with high level instructions for coding agents

juliandescottes:
Just basic instructions, no emojis, no excessive comments, no co-authored messages.


dmose (approved):
r=dmose

PR #81, opened 2026-05-01
fix: create user home directory in Docker image

alexghr:
This PR creates a home directory for the `nodejs` user inside the container. Firefox tries to create folders inside the home dir and without it (`nodejs` is a system user so it doesn't get a home directory created by default) crashes at runtime.


juliandescottes (approved):
This looks correct, thanks for the fix!

PR #82, opened 2026-05-03
feat: add press_key MCP tool for keyboard input simulation

jasonanovak:
Lets the model send keyboard input to the browser — named keys (Enter, Escape, Tab, arrows, F1–F12), modifier combinations (ctrl+l, ctrl+shift+t, alt+F4), or plain characters — directed at a specific snapshot element (uid) or at the currently focused element.

Implementation:

- dom.ts: add KEY_MAP and MODIFIER_MAP (named keys and modifier aliases mapped to Selenium Key.* unicode constants) and a pressKey(key, uid?) method that parses "ctrl+shift+t"-style combos. With a uid it routes through Key.chord + sendKeys on the element; without one it drives selenium's W3C Actions keyboard source directly so the key goes to the active element.
- firefox/index.ts: expose FirefoxClient.pressKey with a not-connected guard.
- tools/input.ts: define pressKeyTool schema and handlePressKey, with the same UID error mapping used by the other input tools.
- tools/index.ts, src/index.ts: register the tool and handler.

Also drops an orphaned yaml dev-dependency entry from package-lock.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7040,24 +7040,6 @@
         "node": ">=10"
       }
     },
-    "node_modules/yaml": {
-      "version": "2.8.3",
-      "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.8.3.tgz",
-      "integrity": "sha512-AvbaCLOO2Otw/lW5bmh9d/WEdcDFdQp2Z2ZUH3pX9U2ihyUY0nvLv7J6TrWowklRGPYbB/IuIMfYgxaCPg5Bpg==",
-      "dev": true,
-      "license": "ISC",
-      "optional": true,
-      "peer": true,
-      "bin": {
-        "yaml": "bin.mjs"
-      },
-      "engines": {
-        "node": ">= 14.6"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/eemeli"
-      }
-    },
     "node_modules/yargs": {
       "version": "17.7.2",
       "resolved": "https://registry.npmjs.org/yargs/-/yargs-17.7.2.tgz",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mozilla/firefox-devtools-mcp",
-  "version": "0.9.15",
+  "version": "0.10.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@mozilla/firefox-devtools-mcp",
-      "version": "0.9.15",
+      "version": "0.10.1",
       "license": "MIT OR Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "1.29.0",
@@ -287,7 +287,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
@@ -336,7 +335,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       }
@@ -348,7 +346,6 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
-      "peer": true,
       "dependencies": {
         "@emnapi/wasi-threads": "1.2.2",
         "tslib": "^2.4.0"
@@ -361,7 +358,6 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
-      "peer": true,
       "dependencies": {
         "tslib": "^2.4.0"
       }
@@ -2234,7 +2230,6 @@
       "integrity": "sha512-rLoGZIf9afaRBYsPUMtvkDWykwXwUPL60HebR4JgTI8mxfFe2cQTu3AGitANp4b9B2QlVru6WzjgB2IzJKiCSA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.58.0",
         "@typescript-eslint/types": "8.58.0",
@@ -2446,7 +2441,6 @@
       "integrity": "sha512-lt3kovsyHwYe00wq4D1ti0Z974fWj4NLp6siqiyEufUpyFwK9Yhi7rBhac9JL5aA0zoMrJqc4vYPZRUnI7l7nw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@bcoe/v8-coverage": "^1.0.2",
         "@vitest/utils": "4.1.8",
@@ -2549,7 +2543,6 @@
       "integrity": "sha512-RUS2ZU2TsduVrI+9c12uTNaKrNUTsm6yFt3fueEUB9iKvyC2UP83F+sqIz00HQIah4UOL1TMoDAki8K0NjGvsA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@vitest/utils": "4.1.8",
         "fflate": "^0.8.2",
@@ -2666,7 +2659,6 @@
       "integrity": "sha512-xRQbDb9BnwDafYNn6Vwl839DYVjqXYb1XVGtWAZ1kcDc6iwAL4hg3B1dZlRiuENFeO2H53gFG3in621AdERVAg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -3179,7 +3171,6 @@
       "integrity": "sha512-X7sjQzceUhu1u7Y/ylrRZFU2FS6LRiFVp6rKLPg23y3x3c3DOKAwuXGDp+PAGjh6CSnCjYeAul8pcT8bAl+lSA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "mdn-data": "2.27.1",
         "source-map-js": "^1.2.1"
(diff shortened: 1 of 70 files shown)


src/tools/input.ts
```diff
@@ -307,3 +328,28 @@ export async function handleUploadFileByUid(args: unknown): Promise<McpToolRespo
...
+    if (!key || typeof key !== 'string') {
+      throw new Error('key parameter is required and must be a string');
+    }
+
+    const { getFirefox } = await import('../index.js');
+    const firefox = await getFirefox();
+
+    try {
+      await firefox.pressKey(key, uid);
+      return successResponse(uid ? `✅ press_key "${key}" on ${uid}` : `✅ press_key "${key}"`);
```
juliandescottes, line 345: Remove the emojis from the logs

package-lock.json
juliandescottes, line 1: Is the package-lock update necessary? If not, can you remove it from this changeset?

src/firefox/dom.ts
```diff
@@ -303,6 +352,54 @@ export class DomInteractions {
...
+      if (!this.resolveUid) {
+        throw new Error(
+          'pressKey: resolveUid callback not set. Ensure snapshot is initialized.'
+        );
+      }
+      const el = await this.resolveUid(uid);
+      // Key.chord concatenates modifiers + key + Key.NULL (releases all modifiers)
+      await el.sendKeys(Key.chord(...modifiers, mainKey));
+    } else {
+      // Send to the currently focused element via W3C Actions keyboard source
```
juliandescottes, line 387: Maybe WebDriver instead of W3C ? https://www.w3.org/TR/webdriver2/#actions 

W3C actions is a bit vague.

juliandescottes, line 387: Maybe WebDriver instead of W3C ? https://www.w3.org/TR/webdriver2/#actions 

W3C actions is a bit vague.

src/firefox/dom.ts
```diff
@@ -4,6 +4,55 @@
 
 import { By, Key, WebDriver, WebElement } from 'selenium-webdriver';
 
+// Map of lowercase key names → Selenium unicode values
+const KEY_MAP: Record<string, string> = {
+  enter: Key.RETURN,
+  return: Key.RETURN,
```
juliandescottes, line 10: RETURN and ENTER are mapped to different keys in selenium, I think we should stick to that. https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/selenium-webdriver/lib/input.js#L53-L54

Overall we are missing a few keys, might be good to add them.

src/firefox/dom.ts
```diff
@@ -303,6 +352,54 @@ export class DomInteractions {
...
+  async pressKey(key: string, uid?: string): Promise<void> {
+    const parts = key.split('+').map((p) => p.trim().toLowerCase());
+    const modifiers: string[] = [];
+    let mainKey = '';
+
+    for (const part of parts) {
+      if (part in MODIFIER_MAP) {
+        modifiers.push(MODIFIER_MAP[part]!);
+      } else {
+        mainKey = KEY_MAP[part] ?? part;
```
juliandescottes, line 369: Consider throwing here if `mainKey` was already set. I asked it to press ctrl+k+l, which resulted in only ctrl+l being sent. 

Maybe adapt the tool description as well to mention that you can use many modifiers, but only one key?

juliandescottes (changes requested):
Can you amend the commit to not be co-authored by the agent? 

(I had a PR to add this to AGENTS.md, but I only merged it today sorry!)

Otherwise the feature looks useful. There's a small gap in terms of inputs if you want to type text in a field, you either have fill_by_uid, which overwrites everything. Or we could reuse that new tool, but at the moment it's really meant for single key usage.

One thing to note is that after using the tool a bit, the agent started using it to type text, one command per character. Either we update the description to explain that you can actually send several keys at once, or we make it really clear that this should not be used for typing.

Could you add some tests as well?

juliandescottes:
@jasonanovak we would still be interested to merge this PR if you have time to update it.

In case we don't hear back or if you can't update it, hopefully it's fine if I rebase it and apply my suggestions in a followup.

jasonanovak:
Hi @juliandescottes sorry for the delayed response. Life got nuts. I am going to update the PR now and address your comments.

src/firefox/dom.ts
jasonanovak: Changed the mapping and added the keys that are in selenium.

src/tools/input.ts
jasonanovak: Done

package-lock.json
jasonanovak: Removed it.

src/firefox/dom.ts
jasonanovak: made this webdriver consistently

jasonanovak: made this webdriver consistently

src/firefox/dom.ts
jasonanovak: Clarified.

jasonanovak:
Ok! Made the changes.

Added tests as well.

Finally, tried to make it clear that this wasn't for typing and that's more fill_by_uid.

juliandescottes:
Thanks @jasonanovak ! 

Will take a look today, cc-ing @freema for review as well since he's working on a similar PR.

src/firefox/dom.ts
```diff
@@ -288,6 +416,38 @@ export class DomInteractions {
...
+   */
+  async pressKey(key: string, uid?: string): Promise<void> {
+    const { modifiers, key: mainKey } = parseKeyCombo(key);
+
+    if (uid) {
+      if (!this.resolveUid) {
+        throw new Error('pressKey: resolveUid callback not set. Ensure snapshot is initialized.');
+      }
+      const el = await this.resolveUid(uid);
+      await el.sendKeys(Key.chord(...modifiers, mainKey));
```
juliandescottes: Instead of using `sendKeys` when we have a uid and `actions` in the other case, could we try to focus the element corresponding to the uid and then use actions ?

Otherwise I'm worried we would get slightly different behaviors when using a uid or not.

src/firefox/dom.ts
juliandescottes: I didn't realize that Selenium's Key.ENTER was the numpad one. I'm not sure what's the best approach between being faithful to selenium's Key definition or aliasing `enter` to `Key.RETURN`.

It probably doesn't change much in most cases, but maybe for now we should revert to what you had, and add `numpadenter: Key.ENTER`?

src/tools/input.ts
```diff
@@ -147,6 +147,30 @@ export const uploadFileByUidTool = {
...
+  annotations: {
+    readOnlyHint: false,
+  },
+  inputSchema: {
+    type: 'object',
+    properties: {
+      key: {
+        type: 'string',
+        description:
+          'One key, optionally preceded by "+"-separated modifiers, such as "Escape", "F5" or "ctrl+shift+t". Modifiers: ctrl, alt, shift, meta. Named keys: Enter (numpad), Return, Tab, Backspace, Delete, Insert, Space, Escape, Home, End, PageUp, PageDown, ArrowUp, ArrowDown, ArrowLeft, ArrowRight, F1-F12, Numpad0-Numpad9, Clear, Pause, Help, Cancel, Semicolon, Equals, Add, Subtract, Multiply, Divide, Decimal, Separator. Anything else must be a single character.',
```
juliandescottes: Might be confusing to mention `F5` here. Most readers (and models) would probably interpret that as the "reload" shortcut, but it would only work from Chrome/privileged context, and will be misleading in other situations.

To be safe I would remove it.

juliandescottes (changes requested):
Sorry for the delay! Main comment is that we should try to consistently use actions and stop falling back to sendKey if a uid is passed.

Once the last comments are addressed we should be good to go! 

Thanks

freema (commented):
Thanks for picking this up @jasonanovak, and thanks @juliandescottes for the ping. Ran the branch against Firefox 154 (macOS, headless): lint, typecheck, tests and build green, merges cleanly with main; measurements behind Julian's points are in the threads. Nothing exercises `DomInteractions.pressKey` itself, I can port the integration test from #153 if useful (the README tool overview could list `press_key` too). Once the uid path goes through actions I am happy and will rebase #153 on top so it shrinks to `type_text`.

src/firefox/dom.ts
freema: +1, measured: `Key.chord` appends `Key.NULL` and Element Send Keys dispatches it, so `press_key Escape` with a uid yields Escape plus a phantom `Unidentified` keydown/keyup (actions path: Escape only). On `<input type=file>` it fails with `File not found:`, which `handleUidError` rewrites to "stale uid, take_snapshot", so the model loops. Focus-then-actions fixes both locally; a non-focusable uid then needs its own error, and the caret lands at the start of a contenteditable instead of the end.

src/firefox/dom.ts
freema: Measured: `Return` arrives as `code=Enter`, `Enter` as `code=NumpadEnter`. +1 to `enter -> Key.RETURN` plus `numpadenter: Key.ENTER`.

src/tools/input.ts
freema: `ctrl+shift+t` has the same issue: from content neither triggers the browser action (checked: tab count unchanged, no reload). A page-level example like `shift+Tab` would be safer. On macOS `ctrl+a` also does not select all (`meta+a` does).

juliandescottes:
@jasonanovak quick heads up: since this PR will unblock #153 , I might push an additional commit to address my suggestions so that we can land this sooner rather than later. Please shout if you're planning to get to it soon.

I would prefer if you could author all the changes, but I also don't want to hold up the next PR for too long :)

juliandescottes (approved):

juliandescottes:
@jasonanovak I just pushed an additional commit to address the review comments, and I merged the PR. Thanks a lot for the contributions!

PR #84, opened 2026-05-07
Bug 2037806 - chore: adapt publishing script/workflow for trusted publisher

juliandescottes:
Also slightly updates the script to generate the -moz package


juliandescottes:
Mostly a follow up, merging.

PR #86, opened 2026-05-14
fix: kill zombie geckodriver when Firefox is closed

csamu:
When Firefox is closed while the MCP server is running, the server freezes and never recovers. This is especially frustrating when you accidentally close Firefox windows that the AI agents use. This fix makes it recover automatically. 

## The problem

**TLDR:** The broken geckodriver TCP connection with Firefox is never detected and Selenium has no command timeout. The MCP server waits "forever".

1. A non-headless MCP session is started
2. User closes Firefox (or it crashes)
3. Any MCP tool call arrives (e.g. take_snapshot)
	- `getFirefox()` checks isConnected() which fails because Firefox is gone
	- `getFirefox()` then calls `resetFirefox()` which calls `FirefoxCore.close()` which calls Seleniums `WebDriver.quit()`
	- `WebDriver.quit()` sends HTTP DELETE `/session` to geckodriver
		- geckodriver translates the DELETE into a Marionette command and sends it to Firefox over raw TCP
		- Since Firefox is gone, geckodriver tries to **send commands over a broken TCP connection**, and that blocks indefinitely. Probably a race condition according to AI.
	- `WebDriver.quit()` keeps waiting for the HTTP response from geckodriver which never comes
	- `WebDriver.quit()`  never executes its `onQuit` callback that stops geckodriver.
	- Geckodriver is now zombie
4. The tool call never returns. The entire MCP server is now stuck.

So there are 2 upstream bugs: One in Selenium and one in Geckodriver. If those are fixed, this patch becomes unecessary. But the MCP server can still work even without those fixes.

## The solution

 **TLDR:** Give `FirefoxCore.close()` a 5-second timeout. If it hangs, force-kill geckodriver and tear down the dead session . The next tool call starts a fresh Firefox automatically.
 
The code that should kill geckodriver but is never triggered, the `onQuit` callback, is actually _secretly exposed_ in the Selenium WebDriver object: `WebDriver.onQuit_()`. **Notice the underscore at the end**. If you call this, geckodriver just dies. We use this to bypass the mistake in Selenium.
 
There is no custom kill command. `onQuit_()` uses the internal Selenium `proc.kill('SIGTERM')`, which is cross-platform and has no shell commands.

## Changes

- Add: `FirefoxCore.killService()` -- kills the geckodriver process using selenium's onQuit_()
- Add: `FirefoxClient.killService()` -- exposes killService on the public facade
- Edit: `FirefoxCore.reset()` -- close log file fd, swallow `quit()` rejection
- Edit: `resetFirefox()` -- Add 5s timeout to close, with a fallback to killService() + reset()
- Minor related edits

## Test

`scripts/test-closed-window.js` covers two scenarios:

- **SIGSTOP:** Freezes Firefox so close() genuinely hangs
- **SIGKILL:** Kills Firefox completely

Run test with: `node scripts/test-closed-window.js`

### DIFF
--- a/scripts/test-closed-window.js
+++ b/scripts/test-closed-window.js
@@ -1,0 +1,363 @@
+#!/usr/bin/env node
+/**
+ * Test: zombie geckodriver cleanup
+ *
+ * Two scenarios:
+ *
+ * Scenario A (SIGSTOP): Firefox is frozen and can't respond, so close() hangs.
+ *   Proves that killService() + reset() kills the zombie when close() can't.
+ *
+ * Scenario B (SIGKILL): Firefox is completely dead (user clicked [X]).
+ *   Recovery test: can we clean up and reconnect after Firefox dies?
+ *   Note: SIGKILL closes TCP sockets, so geckodriver often dies with Firefox.
+ *   A true zombie (geckodriver alive but stuck) is tested by Scenario A.
+ */
+import { FirefoxDevTools } from '../dist/index.js';
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+
+const CLOSE_TIMEOUT_MS = 5000;
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function pgrep(pattern) {
+  try {
+    return execFileSync('pgrep', ['-f', pattern], { encoding: 'utf-8' })
+      .trim()
+      .split('\n')
+      .map(Number)
+      .filter((n) => !isNaN(n));
+  } catch {
+    return [];
+  }
+}
+
+function isAlive(pid) {
+  // /proc/pid/stat format: pid (comm) state ...
+  // The comm field can contain spaces, so we find state after the last ')'.
+  try {
+    const stat = readFileSync(`/proc/${pid}/stat`, 'utf-8');
+    const closeParen = stat.lastIndexOf(')');
+    const state = stat[closeParen + 2];
+    return state !== 'Z';
+  } catch {
+    return false;
+  }
+}
+
+function killHard(pid) {
+  try {
+    process.kill(pid, 'SIGKILL');
+  } catch {}
+}
+
+function freeze(pid) {
+  try {
+    process.kill(pid, 'SIGSTOP');
+  } catch {}
+}
+
+function getDescendants(parentPid) {
+  const result = [];
+  const queue = [parentPid];
+  while (queue.length > 0) {
+    const pid = queue.shift();
+    try {
+      const output = execFileSync('pgrep', ['-P', String(pid)], {
+        encoding: 'utf-8',
+      });
+      const children = output
+        .trim()
+        .split('\n')
+        .map(Number)
+        .filter((n) => !isNaN(n));
+      result.push(...children);
+      queue.push(...children);
+    } catch {
+      // No children
+    }
+  }
+  return result;
+}
+
+function waitForDeath(pid, timeoutMs) {
+  return new Promise((resolve) => {
+    const start = Date.now();
+    const check = () => {
+      if (!isAlive(pid)) {
+        resolve(true);
+      } else if (Date.now() - start > timeoutMs) {
+        resolve(false);
+      } else {
+        setTimeout(check, 100);
+      }
+    };
+    check();
+  });
+}
+
+function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function verifyFrozen(pids) {
+  for (const pid of pids) {
+    if (!isAlive(pid)) continue;
+    try {
+      const stat = readFileSync(`/proc/${pid}/stat`, 'utf-8');
+      const closeParen = stat.lastIndexOf(')');
+      const state = stat[closeParen + 2];
+      if (state !== 'T') {
+        return { ok: false, pid, state };
+      }
+    } catch {
+      // Process died — that's fine
+    }
+  }
+  return { ok: true };
+}
+
+// Suppress unhandled rejections from fire-and-forget promises
+process.on('unhandledRejection', () => {});
+
+// ---------------------------------------------------------------------------
+// Launch helper — creates FirefoxDevTools, connects, returns instance + PIDs
+// ---------------------------------------------------------------------------
+
+async function launchFirefox(geckosBefore, excludePids = []) {
+  const devTools = new FirefoxDevTools({
+    headless: true,
+    viewport: { width: 1280, height: 720 },
+  });
+  await devTools.connect();
+
+  const geckoPid = pgrep('geckodriver').find(
+    (p) => !geckosBefore.has(p) && !excludePids.includes(p)
+  );
+  if (!geckoPid) {
+    throw new Error('No geckodriver PID found after connect');
+  }
+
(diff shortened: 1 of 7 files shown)


csamu:
I updated the PR text for better understanding.

src/index.ts
```diff
@@ -41,11 +41,39 @@ let nextLaunchOptions: FirefoxLaunchOptions | null = null;
 let pendingWarning: string | null = null;
 
 /**
- * Reset Firefox instance (used when disconnection is detected)
+ * Reset Firefox instance (used when disconnection is detected).
+ * Tries graceful close with a timeout; if it hangs (zombie geckodriver),
+ * force-kills the process tree.
  */
-export function resetFirefox(): void {
+export async function resetFirefox(): Promise<void> {
```
juliandescottes, line 48: I think this method should remain simple. 
Let's fix this entirely in firefox/core.ts. It already contains mentions about onQuit_ and selenium quit issues so I think it's the right spot to add yet another one.

src/firefox/core.ts
```diff
@@ -255,14 +255,36 @@ export class FirefoxCore {
...
   }
 
+  /**
+   * Kill the geckodriver service process using selenium's own service.kill().
+   * Cross-platform — no shell commands needed.
+   * Used when graceful close() hangs (e.g. Firefox died but geckodriver didn't notice).
+   */
+  killService(): void {
+    if (this.driver) {
+      void (this.driver as any).onQuit_().catch(() => {});
```
juliandescottes: We should check if onQuit_ is available before, it's an optional argument on the webdriver side if I remember correctly.

src/firefox/core.ts
```diff
@@ -255,14 +255,36 @@ export class FirefoxCore {
...
     this.currentContextId = null;
+
+    if (this.logFileFd !== undefined) {
+      try {
+        closeSync(this.logFileFd);
+      } catch {
+        // already closed
+      }
+      this.logFileFd = undefined;
+    }
```
juliandescottes: This is duplicated from close(), I imagine you had issues in the next connect() call when trying to open the same file again? 

At this point it seems like reset() and close() should apply the same logic and could be merged. For instance, `reset` is currently not resetting env variables, but that's most likely a bug which means on the next connect() the "custom" env variables will now be considered as original ones.

juliandescottes (changes requested):
Thanks for the patch. Let's try to keep the complexity in firefox/core.ts

juliandescottes:
Will review today, thanks

src/firefox/core.ts
```diff
@@ -384,27 +365,60 @@ export class FirefoxCore {
...
+      try {
+        webdriver._bidiConnection.close();
+      } catch {
+        /* already dead */
       }
-      // In connect-existing mode, geckodriver's DELETE /session releases Marionette
-      // without terminating Firefox (since geckodriver was started with --connect-existing).
-      if ('quit' in this.driver) {
-        await (this.driver as { quit(): Promise<void> }).quit();
+      webdriver._bidiConnection = undefined;
```
```suggestion
      try {
        webdriver._bidiConnection.close();
      } finally {
        webdriver._bidiConnection = undefined;
      }
```

Use try / finally here?

src/firefox/index.ts
```diff
@@ -465,24 +465,21 @@ export class FirefoxClient {
...
   // ============================================================================
   // Cleanup
   // ============================================================================
 
   async close(): Promise<void> {
-    await this.core.close();
+    try {
+      await this.core.close();
+    } catch {
+      /* never throws per contract */
```
juliandescottes: This comment is a bit vague. Which contract is this referring to? Is `firefox.close` never supposed to throw? 

It's fine to swallow failures here, but let's drop the confusing comment and instead log something to note that `close` failed.

src/firefox/core.ts
```diff
@@ -384,27 +365,60 @@ export class FirefoxCore {
...
+        await Promise.race([
+          (webdriver as { quit(): Promise<void> }).quit(),
+          new Promise<never>((_, reject) => {
+            timer = setTimeout(() => reject(new Error('close timeout')), webdriverQuitTimeout);
+          }),
+        ]);
+      } catch {
+        // Timer has passed but webdriver.quit() still hasn't returned.
+        // This means Geckodriver is unresponsive. Kill geckodriver.
+        // Calling webdriver.onQuit_ (notice the underscore at the end) kills the geckodriver process
```
juliandescottes: We also hit this codepath if `quit()` throws/rejects early (and one of the tests you added exercises that). Let's simplify the comment.
```suggestion
        // webdriver.quit() either failed or timed out.
        // Call webdriver.onQuit_ to kill geckodriver.
```

(edit: I just noticed you have a logDebug below, it might be self-explanatory enough thanks to this... feel free to completely remove the comment)

scripts/test-closed-window.js
```diff
@@ -0,0 +1,270 @@
...
+
+async function main() {
+  console.log('--- Zombie geckodriver fix test ---\n');
+  const geckosBefore = new Set(pgrep('geckodriver'));
+  const usedPids = [];
+
+  // ---------------------------------------------------------------
+  // Scenario A: Firefox frozen (SIGSTOP)
+  // ---------------------------------------------------------------
+  console.log('Scenario A: Firefox frozen (SIGSTOP)');
```
juliandescottes: The comment just repeats the content of the log, let's remove it. Same comment on the rest of the file.

scripts/test-closed-window.js
```diff
@@ -0,0 +1,270 @@
...
+
+  console.log('  3. Calling close() (built-in timeout + force-kill)...');
+  const t0 = Date.now();
+  await a.devTools.close();
+  console.log(`     close() completed in ${Date.now() - t0}ms`);
+
+  // Verify geckodriver death BEFORE cleaning up frozen Firefox (avoids false positives)
+  console.log('  4. Verifying geckodriver was killed by close()...');
+  if (!(await waitForDeath(a.geckoPid, 5000))) {
+    console.log('  [FAIL] Geckodriver still alive after close()');
```
juliandescottes: Let's use `console.error` for failures, same comment for the other [FAIL] logs in the file.

scripts/test-closed-window.js
```diff
@@ -0,0 +1,270 @@
...
+    killAll([...a.firefoxPids, a.geckoPid]);
+    process.exit(1);
+  }
+  console.log('     Geckodriver is dead');
+
+  console.log('  5. Cleaning up frozen Firefox...');
+  killAll(a.firefoxPids);
+
+  console.log('  6. Reconnecting...');
+  usedPids.push(a.geckoPid, await reconnect(geckosBefore, usedPids));
```
juliandescottes: If the intent here is to have `a.geckoPid` in `usedPids` for the `reconnect` call, we should split it on 2 lines
```suggestion
  usedPids.push(a.geckoPid);
  usedPids.push(await reconnect(geckosBefore, usedPids));
```

Same comment for the other similar spots.

juliandescottes (changes requested):
A few comments to address, especially on the test, but we should be close.

csamu:
I changed according to feedback. 
I also discovered that the `SIGSTOP` test scenario was bogus an never worked, so I removed it from `test-closed-window.js`, but it doesn't affect the core PR issue where Firefox is closed.

juliandescottes (approved):
Thanks, looks good to me!

PR #93, opened 2026-06-04
chore: prefix all package references with @mozilla in README.md

juliandescottes:


LukasHeimann:
I think the `npx ...` commands in the readme will also need to be updated to point to the prefix now.

juliandescottes:
> I think the `npx ...` commands in the readme will also need to be updated to point to the prefix now.

Good catch, thanks!

PR #94, opened 2026-06-11
Bug 2046777 - fix: use about:blank as start url instead of about:home

juliandescottes:
about:home is restricted and can lead to issues with recent builds of Firefox Nightly.


juliandescottes:
I'm going to go ahead and get that landed, this is potentially quite annoying for internal usage.

PR #96, opened 2026-06-11
Bug 2046888 - feature: Add --log-file support

juliandescottes:
This PR adds a new cli argument `--log-file` which can be used to save the logs on to a specified file path, to make it easier to debug past sessions.


padenot (approved):
r+ of course. I'll never say no to a patch adding ways to log.

PR #98, opened 2026-06-12
Bug 2018268 - fix: migrate script evaluation tools to WebDriver BiDi  

juliandescottes:
In preparation for the debugging tool, we need to switch all script evaluation to BiDi. Main reason is that BiDi will automatically pick up the paused frame for script evaluation if you are stopped at a breakpoint, but WebDriver Classic will not.

Also taking the opportunity to align both the privileged script evaluation with the regular one. It frankly doesn't need a dedicated tool/command, but for now let's keep it as is.

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -1,8 +1,9 @@
 /**
- * JavaScript evaluation tool (currently disabled - see docs/future-features.md)
+ * JavaScript evaluation tool
  */
 
 import { successResponse, errorResponse } from '../utils/response-helpers.js';
+import { remoteValueToNative } from '../utils/remote-value.js';
 import type { McpToolResponse } from '../types/common.js';
 
 export const evaluateScriptTool = {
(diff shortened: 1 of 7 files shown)


src/tools/script.ts
```diff
@@ -130,42 +98,51 @@ export async function handleEvaluateScript(args: unknown): Promise<McpToolRespon
...
-      const timeoutValue = (args as { timeout?: number })?.timeout ?? DEFAULT_TIMEOUT;
+    const callFunctionPromise = firefox.sendBiDiCommand('script.callFunction', {
+      functionDeclaration: fnString,
+      awaitPromise: true,
+      arguments: resolvedArgs,
+      target: { context: firefox.getCurrentContextId() },
+    });
+
+    // Race against timeout as WebDriver BiDi callFunction has no built-in
+    // timeout feature.
```
padenot, line 110: Looks like this is https://www.w3.org/TR/webdriver-bidi/#issue-6eade172

padenot (approved):

src/tools/script.ts
juliandescottes: Exactly, the topic hasn't moved much since we landed the initial script commands (to be clear, evaluate has the same issue, but it has a clearer TODO in the spec :) )

PR #100, opened 2026-06-18
Bug 2047657 - feature: add support for debugging logpoints

juliandescottes:
At the moment full breakpoint support is not easy to integrate with the mcp, because it breaks many tools. Instead having logpoints allows the mcp to collect information while running a scenario without breaking the execution.

Ideally the logpoint feature would be built in webdriver-bidi but for the time being we prototype the feature in the MCP layer.

The MCP will set breakpoints and send script.evaluate commands to capture logpoint data, which can be queried out of band by the client.


juliandescottes:
Adding @lutien to review for the WebDriver BiDi side.

lutien (approved):

padenot (approved):

PR #101, opened 2026-06-18
Bug 2047607 - fix: allow to set remote.log.level from --pref

juliandescottes:
This is a known limitation from geckodriver, which has to be able to drive the log level independently and does that by setting the preference `remote.log.level`.

The patch checks if `remote.log.level` is present in the user-provided preferences, and if so, translates the value to a geckodriver log level.

### DIFF
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -244,6 +244,11 @@ export class FirefoxCore {
         log(`Capturing Firefox output to: ${this.logFilePath}`);
       }
 
+      const remoteLogLevel = this.options.prefs?.['remote.log.level'];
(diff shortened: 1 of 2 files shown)


juliandescottes:
Alternatively we could also introduce a dedicated command line argument (eg `--geckodriver-log-level`), but I feel like leaning into the existing cli arguments supported is better than pushing extra complexity towards users.

That being said, it implies users have to know about `remote.log.level`, I guess I should at least add it in the documentation.

juliandescottes:
Tagging @padenot for review, since this is about getting better logs :)

padenot (approved):

PR #121, opened 2026-07-09
Bug 2053928 - feat: add mcpb build target

juliandescottes:


scripts/build-mcpb.mjs
```diff
@@ -0,0 +1,60 @@
...
+  const stagedPkg = {
+    name: pkg.name,
+    version: pkg.version,
+    private: true,
+    dependencies: pkg.dependencies,
+  };
+  writeFileSync(resolve(stagingDir, 'package.json'), JSON.stringify(stagedPkg, null, 2) + '\n');
+
+  console.log('Installing production dependencies...');
+  execSync('npm install --omit=dev', { cwd: stagingDir, stdio: 'inherit' });
```
hbenl: We could make this more predictable/reproducible by also copying `package-lock.json`. And we could use `npm ci` instead of `npm install`.

hbenl (approved):

scripts/build-mcpb.mjs
juliandescottes: Good idea, will update

PR #126, opened 2026-07-20
Bug 2056469, Bug 2056470 - fix connect-existing docs and throw explicit error for sessions without BiDi

matijaerceg:
## Summary

Repurposed per the review discussion. Two changes:

1. README and CLI help: connect-existing now documents starting Firefox with both `--marionette` and `--remote-debugging-port`, and no longer claims console/network events are unavailable in this mode.
2. When `--connect-existing` attaches to a session without a `webSocketUrl` capability, the server fails the connect with an error saying which flags to relaunch with, instead of tools failing later with `Cannot read properties of undefined (reading 'replace')`.

The classic-navigation fallback from the first version of this PR is gone.

## Testing

- typecheck, lint, build, and the vitest suite pass; added unit tests for the new check (missing capability, present capability, launch mode untouched).
- Manually verified on Windows 11 / Firefox 152.0.6 against the built server over stdio: with both flags all tools work; with `--marionette` alone every tool call now fails immediately with the new message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,19 +141,17 @@ Port forwarding between the host and device is handled automatically by geckodri
 
 ### Connect to existing Firefox
 
-Use `--connect-existing` to automate your real browsing session — with cookies, logins, and open tabs intact:
+Use `--connect-existing` to automate your real browsing session, with cookies, logins, and open tabs intact:
 
 ```bash
-# Start Firefox with Marionette enabled
-firefox --marionette
+# Start Firefox with Marionette and the Remote Agent (BiDi)
+firefox --marionette --remote-debugging-port
 
 # Run the MCP server
 npx @mozilla/firefox-devtools-mcp --connect-existing --marionette-port 2828
 ```
 
-Or set `marionette.enabled` to `true` in `about:config` (or `user.js`) to enable Marionette on every launch.
-
-BiDi-dependent features (console events, network events) are not available in connect-existing mode; all other features work normally.
+Both flags are required. `--marionette` lets the MCP attach to the running Firefox, and `--remote-debugging-port` starts the Remote Agent that provides the WebDriver BiDi endpoint used by navigation, console, and network tools. Setting `marionette.enabled` in `about:config` or `user.js` is not sufficient on its own, because the Remote Agent only starts when the command line flag is passed. If Firefox is running without it, the MCP server fails to connect and asks you to restart Firefox with both flags.
 
 > **Warning:** Do not leave Marionette enabled during normal browsing. It sets
 > `navigator.webdriver = true` and changes other browser fingerprint signals,
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ firefox --marionette --remote-debugging-port
 npx @mozilla/firefox-devtools-mcp --connect-existing --marionette-port 2828
 ```
 
-Both flags are required. `--marionette` lets the MCP attach to the running Firefox, and `--remote-debugging-port` starts the Remote Agent that provides the WebDriver BiDi endpoint used by navigation, console, and network tools. Setting `marionette.enabled` in `about:config` or `user.js` is not sufficient on its own, because the Remote Agent only starts when the command line flag is passed. If Firefox is running without it, the MCP server fails to connect and asks you to restart Firefox with both flags.
+Both flags are required because the MCP uses both WebDriver Classic (`--marionette`) and WebDriver BiDi (`--remote-debugging-port`). If Firefox is only started with `--marionette`, the MCP server fails to connect and asks you to restart Firefox with both flags.
 
 > **Warning:** Do not leave Marionette enabled during normal browsing. It sets
 > `navigator.webdriver = true` and changes other browser fingerprint signals,
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -127,7 +127,7 @@ export const cliOptions = {
   connectExisting: {
     type: 'boolean',
     description:
-      'Connect to an already-running Firefox instance via Marionette instead of launching a new one. Requires Firefox to be running with marionette.enabled=true (set in user.js or launched with --marionette).',
+      'Connect to an already-running Firefox instance via Marionette instead of launching a new one. Firefox must be started with both --marionette and --remote-debugging-port; the BiDi endpoint is required for navigation, console, and network tools.',
     default: (process.env.CONNECT_EXISTING ?? 'false') === 'true',
   },
   marionettePort: {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -127,7 +127,7 @@ export const cliOptions = {
   connectExisting: {
     type: 'boolean',
     description:
-      'Connect to an already-running Firefox instance via Marionette instead of launching a new one. Firefox must be started with both --marionette and --remote-debugging-port; the BiDi endpoint is required for navigation, console, and network tools.',
+      'Connect to an already-running Firefox instance via Marionette instead of launching a new one. Firefox must be started with both --marionette and --remote-debugging-port.',
     default: (process.env.CONNECT_EXISTING ?? 'false') === 'true',
   },
   marionettePort: {
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -279,6 +279,19 @@ export class FirefoxCore {
     this.firefoxVersion = (driverCapabilities.get('browserVersion') as string) ?? null;
     logDebug(`Browser version: ${this.firefoxVersion}`);
 
+    // In connect-existing mode, geckodriver can attach to a Firefox that was
+    // started with --marionette but without --remote-debugging-port. Marionette
+    // works, but the session has no BiDi endpoint, so most tools would later
+    // fail with confusing errors. Detect this early with an actionable message.
+    if (this.options.connectExisting && !driverCapabilities.get('webSocketUrl')) {
+      throw new Error(
+        'Connected to Firefox via Marionette, but the session has no WebDriver BiDi endpoint ' +
+          '(missing webSocketUrl capability). Restart Firefox with both flags: ' +
+          'firefox --marionette --remote-debugging-port. ' +
+          'Navigation, console, and network tools require BiDi.'
+      );
+    }
+
     // Remember current window handle (browsing context)
     this.currentContextId = await this.driver.getWindowHandle();
     logDebug(`Browsing context ID: ${this.currentContextId}`);
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -287,8 +287,7 @@ export class FirefoxCore {
       throw new Error(
         'Connected to Firefox via Marionette, but the session has no WebDriver BiDi endpoint ' +
           '(missing webSocketUrl capability). Restart Firefox with both flags: ' +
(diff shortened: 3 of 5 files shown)


juliandescottes:
Hi @matijaerceg !

`--connect-existing` is not actually marionette only. If you start the browser with `--marionette` + `--remote-debugging-port` you should still be able to use WebDriver BiDi features.

However I can see that our documentation doesn't explain that at https://github.com/mozilla/firefox-devtools-mcp#connect-to-existing-firefox . I think we should fix the documentation, and throw an explicit error when connecting to a non-BiDi firefox instance. Console and network events should also work in connect-existing mode.

My plan is to move the MCP to only use BiDi APIs https://bugzilla.mozilla.org/show_bug.cgi?id=2037070 (because the Classic + BiDi mix is the root cause of many issues).

Can you check if you still have the issue if you use `firefox --marionette --remote-debugging-port` ?

juliandescottes (changes requested):
See question in comment, can you test with `firefox --marionette --remote-debugging-port` ?

matijaerceg:
Confirmed — you're right. Tested on Windows 11 / Firefox 152.0.6 with `firefox --marionette --remote-debugging-port`: the session gets a real `webSocketUrl` and `navigate_page` / `new_page` work on unpatched 0.9.12. The failure only occurs with `--marionette` alone, which is what the README's connect-existing section currently shows (and it also says console/network events don't work in this mode, when they do with the Remote Agent running).

Given the BiDi-only direction in bug 2037070, a silent classic fallback is probably the wrong shape. Happy to rework this PR into what you described: update the README (start Firefox with both flags) and throw an explicit error at connect time when `webSocketUrl` is absent, pointing at `--remote-debugging-port`. Or if you'd rather fold it into the BiDi-only work, feel free to close this — either is fine by me.

juliandescottes:
> Confirmed — you're right. Tested on Windows 11 / Firefox 152.0.6 with `firefox --marionette --remote-debugging-port`: the session gets a real `webSocketUrl` and `navigate_page` / `new_page` work on unpatched 0.9.12. The failure only occurs with `--marionette` alone, which is what the README's connect-existing section currently shows (and it also says console/network events don't work in this mode, when they do with the Remote Agent running).
> 

Great thanks for checking.

> Given the BiDi-only direction in bug 2037070, a silent classic fallback is probably the wrong shape. Happy to rework this PR into what you described: update the README (start Firefox with both flags) and throw an explicit error at connect time when `webSocketUrl` is absent, pointing at `--remote-debugging-port`. Or if you'd rather fold it into the BiDi-only work, feel free to close this — either is fine by me.

Repurposing this sounds good! Let me know if you need any pointers.

juliandescottes:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=2056469 & https://bugzilla.mozilla.org/show_bug.cgi?id=2056470 on bugzilla for tracking both aspects (docs and error) but it's fine to handle both in a single PR. As you prefer!

matijaerceg:
Reworked as discussed: README and the CLI help now require both flags, and connect() fails fast with a pointer to `--remote-debugging-port` when the session has no `webSocketUrl`. Commits reference bug 2056469 and 2056470.

One thing I hit while testing the flag matrix: if Firefox is running without `--marionette` at all, geckodriver retries the Marionette port for about 60 seconds and then fails the session request with `{"error":"timeout","message":"unknown error"}`, so every tool call stalls a minute and reports nothing useful. Happy to add a commit here or do a follow-up that catches that and points at `--marionette`, whichever you prefer, or leave it for the BiDi-only rework.

Unrelated: the vitest suite has 5 pre-existing failures on Windows on main, all path separator assertions in profile.test.ts and core.test.ts; can file that separately if useful.

sleetdrop:
A runtime detail that may not be covered by the PR description, for test reference.

The PR notes tools "fail later with `Cannot read properties of undefined (reading 'replace')`". What I observed on macOS / Firefox 153.0b13 (connect-existing, `--marionette` without `--remote-debugging-port`) is a slightly different failure mode: `driver.getBidi()` (selenium-webdriver `lib/webdriver.js`) receives an `undefined` webSocketUrl and builds a half-initialised connection whose `socket.readyState` stays at `0` (CONNECTING). The subsequent `bidi.subscribe('log.entryAdded')` then **never resolves** (no exception thrown), so `connect()` hangs until the MCP client times out and kills the server (exit status 1) — the `undefined.replace` error never surfaces.

So the real failure is a silent hang, not a thrown `undefined.replace`. This PR checks and throws right after `getCapabilities()` and before `subscribe()`, which **already prevents the hang** — good. Flagging it only because unit tests that mock `getCapabilities()` cover the throw path; the actual subscribe-stage hang depends on how selenium's `getBidi()` handles a missing webSocketUrl, worth keeping in mind for regression.

For my own workflow I keep a working fork at https://github.com/sleetdrop/firefox-devtools-mcp that degrades gracefully instead of failing fast — just to keep things usable while waiting for the proper fix in the right direction (bug 2037070) to land.

(Investigation assisted by a coding agent. macOS arm64, Firefox Developer Edition 153.0b13, firefox-devtools-mcp 0.9.12, connect-existing mode.)

matijaerceg:
Thanks for testing, and glad the check placement covers your case too.

One thing doesn't line up for me, though. The selenium-webdriver that 0.9.12 ships (4.36.0) does this in `getBidi()` (`lib/webdriver.js`):

```js
const caps = await this.getCapabilities()
let WebSocketUrl = caps['map_'].get('webSocketUrl')
this._bidiConnection = new BIDI(WebSocketUrl.replace('localhost', '127.0.0.1'))
```

With no `webSocketUrl` capability, `.replace()` on `undefined` throws a TypeError synchronously — `new BIDI(...)` is never reached, so a half-initialised connection with `socket.readyState === 0` shouldn't be reachable from this path on any platform. That's the `Cannot read properties of undefined (reading 'replace')` I saw on Windows.

The hang symptom you describe (stalls until the MCP client times out and kills the server) matches something else I hit while testing the flag matrix, noted upthread: if the target Firefox doesn't have a working Marionette listener, geckodriver retries the Marionette port for ~60 seconds and then fails the session request. One easy way to end up there is if an already-running Firefox instance grabs the `firefox --marionette ...` invocation, so the flags never apply to the process you're connecting to.

Could you check which selenium-webdriver version your install resolves, and confirm the flags actually took effect (no pre-existing Firefox process, and `about:config` shows the Remote Agent / Marionette state you expect)? Either way, agreed the fail-fast check in this PR sits before any BiDi construction, so both failure modes end at the same explicit error.

juliandescottes:
Still reviewing the code at the moment, but at least for me using Claude Code as client on macos the behavior with the PR looks fine:

> firefox-devtools - navigate_page (MCP)(url: "https://www.mozilla.org")                                                                                                                                                                       
  ⎿  Error: Connected to Firefox via Marionette, but the session has no WebDriver BiDi endpoint (missing webSocketUrl capability). Restart Firefox with both flags: firefox --marionette --remote-debugging-port. Navigation, console, and     
     network tools require BiDi.                                                                                                                                                                                                               
>                                                                                                                                                                                                                                               
> Firefox is running but not with the BiDi endpoint enabled. To fix this, restart Firefox with the required flags:                                                                                                                             
>                                                                                                                                                                                                                                               
>   firefox --marionette --remote-debugging-port 9222                                                                                                                                                                                            
>                                                                                                                                                                                                                                               
>  Or if you want to use the MCP tool to restart:                                                                                                                                                                                               
>                                                                                                                                                                                                                                               
>   firefox-devtools - restart_firefox (MCP)(startUrl: "https://www.mozilla.org")

src/firefox/core.ts
```diff
@@ -279,6 +279,19 @@ export class FirefoxCore {
...
+    // In connect-existing mode, geckodriver can attach to a Firefox that was
+    // started with --marionette but without --remote-debugging-port. Marionette
+    // works, but the session has no BiDi endpoint, so most tools would later
+    // fail with confusing errors. Detect this early with an actionable message.
+    if (this.options.connectExisting && !driverCapabilities.get('webSocketUrl')) {
+      throw new Error(
+        'Connected to Firefox via Marionette, but the session has no WebDriver BiDi endpoint ' +
+          '(missing webSocketUrl capability). Restart Firefox with both flags: ' +
+          'firefox --marionette --remote-debugging-port. ' +
+          'Navigation, console, and network tools require BiDi.'
```
juliandescottes: Let's drop this line, it might get outdated soon.

src/cli.ts
```diff
@@ -127,7 +127,7 @@ export const cliOptions = {
   connectExisting: {
     type: 'boolean',
     description:
-      'Connect to an already-running Firefox instance via Marionette instead of launching a new one. Requires Firefox to be running with marionette.enabled=true (set in user.js or launched with --marionette).',
+      'Connect to an already-running Firefox instance via Marionette instead of launching a new one. Firefox must be started with both --marionette and --remote-debugging-port; the BiDi endpoint is required for navigation, console, and network tools.',
```
juliandescottes: Same comment here, no need to list which tools currently require BiDi under the hood since we fail the connection anyway.

README.md
```diff
@@ -141,19 +141,17 @@ Port forwarding between the host and device is handled automatically by geckodri
...
+firefox --marionette --remote-debugging-port
 
 # Run the MCP server
 npx @mozilla/firefox-devtools-mcp --connect-existing --marionette-port 2828
 ```
 
-Or set `marionette.enabled` to `true` in `about:config` (or `user.js`) to enable Marionette on every launch.
-
-BiDi-dependent features (console events, network events) are not available in connect-existing mode; all other features work normally.
+Both flags are required. `--marionette` lets the MCP attach to the running Firefox, and `--remote-debugging-port` starts the Remote Agent that provides the WebDriver BiDi endpoint used by navigation, console, and network tools. Setting `marionette.enabled` in `about:config` or `user.js` is not sufficient on its own, because the Remote Agent only starts when the command line flag is passed. If Firefox is running without it, the MCP server fails to connect and asks you to restart Firefox with both flags.
```
juliandescottes: Hmmm I think the marionette.enabled preference was a complete hallucination, which I missed to remove. This preference doesn't exist as far as I know, let's drop the whole section about it. And again, no need to explain which tools current use webdriver BiDi.
```suggestion
Both flags are required because the MCP uses both WebDriver Classic (`--marionette`) and WebDriver BiDi (`--remote-debugging-port`). If Firefox is only started with `--marionette`, the MCP server fails to connect and asks you to restart Firefox with both flags.
```

juliandescottes (changes requested):
Great improvement over the current situation, thanks.

Mostly comments about logs/messages, should be straightforward to address.

Also in src/firefox/index.ts we have a few try/catch blocks around BiDi APIs which log 

> logDebug('Debugging events unavailable (BiDi not supported by this Firefox session)');

We should reword the messages. We can keep them to be on the safe side, but say something like "Unable to subscribe to (network/console/debugging) events".

matijaerceg:
Addressed all four: dropped the tool lists from the connect error and CLI help, took your suggested README paragraph as-is, and reworded the three subscribe logs in src/firefox/index.ts to "Unable to subscribe to console/network/debugging events" (try/catch kept). Ready for another look.

juliandescottes (approved):
Thanks @matijaerceg , works for me!

PR #130, opened 2026-07-25
Bug 2057728 - Add saveTo and preview parameters to bulky-output tools

Majdoddin:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2057728

Large tool results currently either bloat the client's context window or get
truncated with data loss; this extends the accepted screenshot saveTo pattern
so the full data goes to disk instead (details and motivation in the bug).

Implements the proposal from the bug: optional saveTo (file path, existing
directory, or true) and preview (character count) on the six bulky-output
tools. The saved file always holds the full untruncated data; the existing
inline truncation limits are unchanged, and responses without saveTo are
identical to before. The screenshot tools are migrated onto the same shared
save helper, with their saveTo widened to also accept true or a directory.

Unit tests cover the save helper and each tool's save/preview paths.

### DIFF
--- a/src/tools/console.ts
+++ b/src/tools/console.ts
@@ -9,6 +9,7 @@ import {
   TOKEN_LIMITS,
   truncateText,
 } from '../utils/response-helpers.js';
+import { saveOutput } from '../utils/save-output.js';
 import type { McpToolResponse } from '../types/common.js';
 
 export const listConsoleMessagesTool = {
@@ -46,6 +47,16 @@ export const listConsoleMessagesTool = {
         enum: ['text', 'json'],
         description: 'Output format (default: text)',
       },
+      saveTo: {
+        type: ['boolean', 'string'],
+        description:
+          'Save all matching messages to a file in full (no truncation or limit) instead of returning them inline. Pass a file path, an existing directory (generated file inside), or true (generated file under ~/.firefox-devtools-mcp/output/). Relative paths resolve against the current working directory.',
+      },
+      preview: {
+        type: 'number',
+        description:
+          'Number of characters of the saved output to return inline as a preview when saveTo is used. Omit for no preview.',
+      },
     },
   },
 };
@@ -64,6 +75,18 @@ export const clearConsoleMessagesTool = {
 
 const DEFAULT_LIMIT = 50;
 
+function formatMessageLine(msg: {
+  level: string;
+  text: string;
+  source?: string | undefined;
+  timestamp?: number | undefined;
+}): string {
+  const timestamp = msg.timestamp ? new Date(msg.timestamp).toISOString() : '';
+  const source = msg.source ? ` [${msg.source}]` : '';
+  const time = timestamp ? `[${timestamp}] ` : '';
+  return `${time}${msg.level.toUpperCase()}${source}: ${msg.text}`;
+}
+
 export async function handleListConsoleMessages(args: unknown): Promise<McpToolResponse> {
   try {
     const {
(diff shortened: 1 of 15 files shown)


juliandescottes:
Hi @Majdoddin thanks for filing a bug and sending a PR. I think it makes sense.

I am working on a similar topic for https://github.com/mozilla/firefox-devtools-mcp/pull/131, but mostly because I would like to prevent writing absolute paths by default. That feels overall risky and error prone, we don't want an agent erasing random files on the user's machine (with a cli param to bypass this if you are fine with the risk).

I was not sure what to pick between `cwd` and `~/.firefox-devtools-mcp`. I ended up picking `~/.firefox-devtools-mcp` because in my scenarios, I usually don't want debugging artifacts polluting my work folder. But I guess your approach makes sense because users will expect relative paths to be relative to cwd.

Anyway, I propose we merge your PR first and I will rebase / adapt mine on top of that (and most likely I will keep "relative" -> cwd and "absolute" -> only accepted if under ~/.firefox-devtools-mcp)

src/tools/console.ts
```diff
@@ -107,6 +134,38 @@ export async function handleListConsoleMessages(args: unknown): Promise<McpToolR
...
+        fileBody,
+        saveTo === true ? undefined : saveTo,
+        'console-messages',
+        format === 'json' ? 'json' : 'txt'
+      );
+      let output = `Console messages (${messages.length} matching, ${totalCount} total) saved to: ${saved.path} (${(saved.bytes / 1024).toFixed(1)}KB)`;
+      if (preview !== undefined && preview > 0) {
+        const previewChars = Math.min(Math.max(preview, 50), TOKEN_LIMITS.MAX_RESPONSE_CHARS);
+        output += '\nPreview:\n' + truncateText(fileBody, previewChars);
+      }
```
juliandescottes, line 165: We could handle this in a single spot rather than duplicating the logic in each tool. 

But since this is explicitly a preview, I don't think we need the full truncation suffix. We could just use `[...]` and add 5 (=length of this short suffix) to the `preview` so that we always respect the requested preview size.

src/tools/network.ts
```diff
@@ -72,6 +75,16 @@ export const listNetworkRequestsTool = {
         enum: ['text', 'json'],
         description: 'Output format (default: text)',
       },
+      saveTo: {
+        type: ['boolean', 'string'],
+        description:
+          'Save all matching requests with full untruncated headers to a file as JSON (ignores limit and detail) instead of returning them inline. Pass a file path, an existing directory (generated file inside), or true (generated file under ~/.firefox-devtools-mcp/output/). Relative paths resolve against the current working directory.',
```
juliandescottes, line 81: I think we should still apply limit and detail when using saveTo, but only if explicitly set. 

We can have up to 1000 requests in memory, so it can be helpful to only select the top N requests. 

For detail we should default to full, but still use an explicit value provided by the user.

src/tools/snapshot.ts
```diff
@@ -43,6 +49,16 @@ export const takeSnapshotTool = {
         type: 'string',
         description: 'CSS selector to scope snapshot to specific element (e.g., "#app")',
       },
+      saveTo: {
+        type: ['boolean', 'string'],
+        description:
+          'Save the complete snapshot text to a file (ignores maxLines) instead of returning it inline. Pass a file path, an existing directory (generated file inside), or true (generated file under ~/.firefox-devtools-mcp/output/). Relative paths resolve against the current working directory.',
```
juliandescottes, line 55: Similar to my comment for network, could we still apply maxLines if explicitly provided?

juliandescottes (approved):
Thanks for the PR! Overall that looks great.

I commented about a few follow up changes, but I will handle them in my PR.

PR #141, opened 2026-07-28
Bug 2058226 - [firefox-devtools-mcp] Add tool to get page text

juliandescottes:
This tool allows to retrieve the page text directly. Before that, asking an agent to summarize or explore a page would usually first start taking a snapshot, and then resort to script evaluation to get the text.

Testing against Claude code, it seems to pickup the get_page_text tool very easily when it needs to get information from the page.

### DIFF
--- a/src/firefox/dom.ts
+++ b/src/firefox/dom.ts
@@ -10,21 +10,6 @@ export class DomInteractions {
     private resolveUid?: (uid: string) => Promise<WebElement>
   ) {}
 
-  /**
-   * Evaluate JavaScript - direct passthrough to executeScript
-   */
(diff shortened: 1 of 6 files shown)


src/firefox/dom.ts
```diff
@@ -10,21 +10,6 @@ export class DomInteractions {
...
-  }
-
-  /**
-   * Get page HTML content
-   */
-  async getContent(): Promise<string> {
-    const html = await this.evaluate('return document.documentElement.outerHTML');
-    return String(html);
-  }
-
```
juliandescottes, line 27: Those webdriver classic commands were only used in an integration test.

hbenl (approved):

PR #144, opened 2026-07-31
Bug 2059668 - Add sandbox option to evaluate_script

Majdoddin:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2059668

evaluate_script always runs in the page's own realm, so inspection is
filtered through whatever the page has overwritten (querySelector, getters,
Array.prototype methods) and anything the function leaves on the global
leaks into the page (details and motivation in the bug).

Implements the proposal from the bug: an optional sandbox string parameter
on evaluate_script, passed through to the BiDi target. The sandbox realm
shares the page DOM behind Xray wrappers with the page's own principal, so
built-ins stay native and writes stay invisible to the page. The same name
reuses the realm across calls; different names are isolated. Element UIDs
passed via args keep working. Without the parameter, the request and
response are identical to before.

Unit tests cover the schema and the BiDi target with and without sandbox.

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -41,6 +41,11 @@ export const evaluateScriptTool = {
         type: 'number',
         description: 'Timeout in ms (default: 5000)',
       },
+      sandbox: {
+        type: 'string',
+        description:
+          'Evaluate in an isolated sandbox realm with this name instead of the page realm. The sandbox shares the page DOM but keeps the native built-ins even where the page overrode them, and its globals stay invisible to the page. The same name reuses the same sandbox across calls; omit to evaluate in the page realm.',
+      },
       saveTo: {
         type: ['boolean', 'string'],
         description:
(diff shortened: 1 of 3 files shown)


src/tools/script.ts
```diff
@@ -41,6 +41,11 @@ export const evaluateScriptTool = {
         type: 'number',
         description: 'Timeout in ms (default: 5000)',
       },
+      sandbox: {
+        type: 'string',
+        description:
+          'Evaluate in an isolated sandbox realm with this name instead of the page realm. The sandbox shares the page DOM but keeps the native built-ins even where the page overrode them, and its globals stay invisible to the page. The same name reuses the same sandbox across calls; omit to evaluate in the page realm.',
```
juliandescottes: I would describe clearly the scope of the isolation.
```suggestion
          'Evaluate in an isolated sandbox realm with this name instead of the page realm. The sandbox shares the page DOM and keeps the native built-ins even where the page overrode them. Page-defined globals and expandos are invisible from the sandbox, and vice-versa. The same name reuses the same sandbox across calls; omit to evaluate in the page realm.',
```

juliandescottes (changes requested):
Thanks! Just one small comment about the description of the options, otherwise this looks fine.

Majdoddin:
Thanks for the review, and for completing the description!

juliandescottes (approved):
Perfect, thanks for the PR!

PR #145, opened 2026-07-31
Bug 2059703 - Make evaluate_privileged_script target an explicit privileged context

Majdoddin:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2059703

evaluate_privileged_script targets the shared "current tab" context id, which
select_page, new_page, and close_page overwrite, so after any tab operation a
privileged eval silently runs in the content page with the page's principal
while still reporting "Script ran in chrome context" (details and history in
the bug).

This makes the target explicit and validated: evaluate_privileged_script now
requires a context parameter, checked against the chrome-scoped
browsingContext.getTree before evaluating - only top-level entries qualify,
since a chrome window's children are content tabs. Invalid or missing targets
fail loudly instead of falling back to whatever tab is current.
select_privileged_context gets the same validation, so it no longer accepts
content ids while claiming a privileged switch. Its Marionette side
(setContext("chrome") and the shared-slot update that the prefs and extension
helpers rely on) is unchanged.

Requiring the parameter is a deliberate breaking change: an implicit default
would reintroduce the hidden state that caused the bug. Explicit per-call
targets are also what WebDriver BiDi itself uses - the protocol has no
current context - and the pattern stateful APIs keep converging on, from
HTTP's stateless requests to Selenium's switchTo() giving way to per-command
context ids in CDP and BiDi. Same direction as bug 2037070.

Unit tests cover the schema, per-call targeting, rejection of content and
unknown ids, the missing-parameter error, the system-access error path, and
select_privileged_context validation.

### DIFF
--- a/src/tools/privileged-context.ts
+++ b/src/tools/privileged-context.ts
@@ -49,7 +49,7 @@ export const selectPrivilegedContextTool = {
 export const evaluatePrivilegedScriptTool = {
   name: 'evaluate_privileged_script',
   description:
-    'Execute JS function in the current privileged context. Requires MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1 env var. Use select_privileged_context first to target a chrome context.',
+    'Execute JS function in a privileged (chrome) browsing context. Requires MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1 env var. Get context ids from list_privileged_contexts.',
   annotations: {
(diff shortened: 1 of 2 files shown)


juliandescottes:
Thanks a lot for the fix. Tested locally, that seems to work fine, and it aligns nicely with what we want to do for the BiDi migration.

juliandescottes (approved):

PR #146, opened 2026-07-31
Typecheck and lint test files

hbenl:


tests/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
```
juliandescottes: I guess we could add this to our main tsconfig.json and remove the --noEmit from the package.json for consistency.

tests/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "rootDir": "..",
+    "paths": {
+      "@/*": ["../src/*"]
+    },
+    "noUncheckedIndexedAccess": false
+  },
+  "include": ["../src/**/*", "**/*"],
```
juliandescottes: Is `"../src/**/*"` needed here since we are already running typechecks for src/**/* via the root tsconfig?

juliandescottes (approved):
Thanks for the cleanup! I completely missed that those were not linted/typechecked. I should have guessed since prettier was already not running on them initially.

juliandescottes:
@hbenl I will let you merge in case you want to update the patch based on one of my comments first, but not blocking of course!

PR #148, opened 2026-08-03
Bug 2060277 - Improve reliablility of element lookups from snapshots

hbenl:


src/firefox/snapshot/injected/treeWalker.ts
juliandescottes: At the moment we have a class of bugs due to this which can become more problematic with this PR.

The short story is that snapshot accepts a `selector` to find a root element. Meaning the root might not be body or html. If somehow the provided root is not "relevant", we will walk the tree, potentially populate the in-content UID map, and ultimately throw because "not relevant". (eg a table with a button, and snapshot for `{selector: "table"}`. 

The "snapshot fails with some selectors" issue was already reported but I never investigated why before this PR. And now since we throw, the "nextElementId" will not be updated in the MCP, and the next snapshot will start assigning the same ids to other elements, and we will end up with duplicated ids.

I would suggest to fix the `isRoot` check here and use `depth == 0`.

But I suspect this is not the only case where the in-content ids might be out of sync with the MCP server (server restarts? bfcache?). The only reason we handle `nextElementId` on the server seems to be to guarantee disjoint ids for different documents? 

What do you think about letting the content page be the only source of truth for the element uids. It can generate a random prefix to avoid cross document clashes, and the MCP server doesn't really need to drive those ids I think?

src/tools/snapshot.ts
```diff
@@ -178,7 +178,7 @@ export async function handleTakeSnapshot(args: unknown): Promise<McpToolResponse
     const displayLines = truncated ? lines.slice(0, maxLines) : lines;
 
     // Build compact output
-    let output = `📸 Snapshot (id=${snapshot.json.snapshotId})`;
+    let output = '📸 Snapshot';
```
juliandescottes: How many emojis did I miss :'( 

Feel free to remove them as you go :)

src/firefox/index.ts
```diff
@@ -60,14 +52,12 @@ export class FirefoxClient {
...
       // Cast to any for BiDi-specific APIs that only exist on selenium WebDriver
       this.consoleEvents = new ConsoleEvents(driver as any, {
-        onNavigate,
         autoClearOnNavigate: false,
       });
 
       this.networkEvents = new NetworkEvents(
         driver as any,
         {
-          onNavigate,
```
juliandescottes, line 70: Should we update console and network events to remove the `onNavigate` support? This was the only call site.

src/firefox/snapshot/manager.ts
```diff
@@ -86,17 +87,17 @@ export class SnapshotManager {
...
-   * Returns text and JSON with snapshotId, no DOM mutations
+   * Returns text and JSON, no DOM mutations
    */
   async takeSnapshot(options?: SnapshotOptions): Promise<Snapshot> {
-    const snapshotId = ++this.currentSnapshotId;
-    this.resolver.setSnapshotId(snapshotId);
-    this.resolver.clear();
-
-    logDebug(`Taking snapshot (ID: ${snapshotId})...`);
+    logDebug('Taking snapshot...');
```
juliandescottes: nit: could be helpful to log the options while we are logging here.

juliandescottes (changes requested):
Thanks Holger! Having stable ids will be really great. See my comment , we have to fix another bug to avoid desyncs with the nextElementId here. 

If possible it would be great to also fully let the page own the ids to avoid more de-syncs (eg bfcache scenarios).

+ a few nits.

hbenl:
addressed the comments in another commit:
- don't register element Uids if creating the snapshot fails
- fix snapshots scoped to an element classified as irrelevant
- log snapshot options
- remove the `onNavigate` option for console and network events
- 🚀 remove another emoji from the code 🎉

juliandescottes (approved):
Alright, looks good to me, thanks @hbenl!

PR #149, opened 2026-08-10
Bug 2060858 - [firefox-devtools-mcp] Add option to let MCP discover accessible Marionette ports

hbenl:
- adds a hidden option `--lookup-marionette-port` and reads the marionette port from the most recent `.port` file written to `~/.firefox-devtools-mcp/instances/`
- adds this option and `--connect-existing` to the MCPB `manifest.json`

I couldn't test this with the MCPB but I tested with the plugin by adding `--connect-existing` and `--lookup-marionette-port` to the `plugin.json` file.

### DIFF
--- a/scripts/build-mcpb.mjs
+++ b/scripts/build-mcpb.mjs
@@ -13,7 +13,7 @@ import { tmpdir } from 'node:os';
 
 const root = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 const pkg = JSON.parse(readFileSync(resolve(root, 'package.json'), 'utf8'));
-const manifest = JSON.parse(readFileSync(resolve(root, 'manifest.json'), 'utf8'));
+const manifest = JSON.parse(readFileSync(resolve(root, 'manifest.mcpb.json'), 'utf8'));
 
 console.log('Building server...');
 execSync('npm run build', { cwd: root, stdio: 'inherit' });
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -4,7 +4,16 @@
 
 import { Builder, Browser, Capabilities, WebDriver } from 'selenium-webdriver';
 import firefox from 'selenium-webdriver/firefox.js';
-import { mkdirSync, openSync, closeSync, existsSync, readdirSync, statSync } from 'node:fs';
+import {
+  mkdirSync,
+  openSync,
+  closeSync,
+  existsSync,
+  readdirSync,
+  statSync,
+  readFileSync,
+} from 'node:fs';
+import { connect as netConnect } from 'node:net';
 import { homedir } from 'node:os';
 import { join, delimiter } from 'node:path';
 import type { FirefoxLaunchOptions } from './types.js';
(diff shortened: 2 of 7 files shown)


scripts/build-mcpb.mjs
```diff
@@ -29,9 +29,13 @@ try {
...
   execSync('npm ci --omit=dev', { cwd: stagingDir, stdio: 'inherit' });
 
+  manifest.version = pkg.version;
+  const args = manifest.server.mcp_config.args.filter(arg => arg !== '--auto-profile');
+  args.push('--connect-existing', '--lookup-marionette-port');
+  manifest.server.mcp_config.args = args;
   writeFileSync(
     resolve(stagingDir, 'manifest.json'),
-    JSON.stringify({ ...manifest, version: pkg.version }, null, 2) + '\n'
+    JSON.stringify(manifest, null, 2) + '\n'
```
juliandescottes: manifest.json is only used to build the mcpb. I guess we could rename it to manifest.mcpb.json for clarity but in any case, you can just add/remove the flags directly there.

src/firefox/core.ts
```diff
@@ -62,6 +70,28 @@ async function findGeckodriverInNpmPackage(): Promise<string | null> {
...
+  }
+  const files = readdirSync(instancesDir);
+  const portFiles = files.filter((f) => /^\d+\.port$/.test(f));
+  const mostRecent = portFiles
+    .map((f) => ({
+      name: f,
+      path: join(instancesDir, f),
+      mtime: statSync(join(instancesDir, f)).mtimeMs,
+    }))
+    .sort((a, b) => b.mtime - a.mtime)[0];
```
juliandescottes, line 88: The name of the port file is supposed to be a PID, could we do a liveness check, in case we have a stale file from a crash?

src/firefox/core.ts
```diff
@@ -62,6 +70,28 @@ async function findGeckodriverInNpmPackage(): Promise<string | null> {
...
+  if (mostRecent) {
+    const port = parseInt(readFileSync(mostRecent.path, 'utf-8').trim(), 10);
```
```suggestion
    const port = Number(readFileSync(mostRecent.path, 'utf-8').trim());
```
in this case we are strictly expecting a number

src/firefox/core.ts
```diff
@@ -136,7 +166,17 @@ export class FirefoxCore {
       const serviceBuilder = new firefox.ServiceBuilder(geckodriverPath);
       this.driver = firefox.Driver.createSession(caps, serviceBuilder.build());
     } else if (this.options.connectExisting) {
-      const port = this.options.marionettePort ?? 2828;
+      let port = this.options.marionettePort ?? 2828;
+      if (this.options.lookupMarionettePort) {
+        const lookedUpPort = lookupMarionettePort();
```
juliandescottes, line 208: Add a logDebug here to say we are looking up a marionette port

src/firefox/core.ts
```diff
@@ -136,7 +166,17 @@ export class FirefoxCore {
...
+        if (lookedUpPort !== undefined) {
+          port = lookedUpPort;
+        } else {
+          throw new Error(
+            'Marionette port not found: please enable Firefox remote control for AI tooling using the AI assistant companion button.'
+          );
+        }
+      }
```
juliandescottes, line 236: A logDebug here as well to say which port we picked up.

src/firefox/core.ts
```diff
@@ -62,6 +70,28 @@ async function findGeckodriverInNpmPackage(): Promise<string | null> {
   }
 }
 
+function lookupMarionettePort(): number | void {
+  const instancesDir = join(homedir(), '.firefox-devtools-mcp', 'instances');
+  if (!existsSync(instancesDir)) {
+    return;
```
juliandescottes, line 78: logDebug here.

src/firefox/core.ts
```diff
@@ -62,6 +70,28 @@ async function findGeckodriverInNpmPackage(): Promise<string | null> {
...
+      name: f,
+      path: join(instancesDir, f),
+      mtime: statSync(join(instancesDir, f)).mtimeMs,
+    }))
+    .sort((a, b) => b.mtime - a.mtime)[0];
+  if (mostRecent) {
+    const port = parseInt(readFileSync(mostRecent.path, 'utf-8').trim(), 10);
+    if (!isNaN(port)) {
+      return port;
+    }
```
juliandescottes, line 101: Let's add a logDebug if we failed to read the port.

src/firefox/core.ts
```diff
@@ -62,6 +70,28 @@ async function findGeckodriverInNpmPackage(): Promise<string | null> {
...
+  const files = readdirSync(instancesDir);
+  const portFiles = files.filter((f) => /^\d+\.port$/.test(f));
+  const mostRecent = portFiles
+    .map((f) => ({
+      name: f,
+      path: join(instancesDir, f),
+      mtime: statSync(join(instancesDir, f)).mtimeMs,
+    }))
+    .sort((a, b) => b.mtime - a.mtime)[0];
+  if (mostRecent) {
```
juliandescottes, line 89: logDebug with the port file path selected.

juliandescottes (changes requested):
Thanks Holger! Looks good overall, will take another look with comments addressed,

src/firefox/core.ts
hbenl: Well, we expect an integer, so why not use `parseInt()`?

src/firefox/core.ts
hbenl: I've added a check that something is listening on the port. This also helps without `--lookup-marionette-port` so I'm using it in this case as well.
I thought checking the PID is less straightforward but apparently you can use `process.kill(pid, 0)` to check if a process exists, I'll try that.

hbenl: `process.kill(pid, 0)` works as expected so I've added that check as well. I'm keeping the port check because I think it's also useful (and the MCP stalls for a minute in geckodriver without that check).

src/firefox/core.ts
juliandescottes: Argh lost my comment. 

Basically, we don't control the folder or its content at all, so we should be strict with what we accept. parseInt will tolerate anything that contains a number, Number is more strict. We can also validate the content with a regexp if you prefer. The goal is just to be more confident we are processing a correct file. Although the liveness check should already help.

src/firefox/core.ts
juliandescottes: yep, process.kill(pid, 0) should be fine here.

juliandescottes (approved):
LGTM, thanks Holger! I will let you decide if you want to keep just parseInt or add some more content validation, won't block on that.

PR #152, opened 2026-08-16
docs: align preset and security docs with the actual tool registry

freema:
## What I noticed

While reading through the tool registry I noticed README.md and SECURITY.md still describe the preset split from before b2e1887 ("fix: Update tool presets based on actual usage"). Every documented preset row is out of date, and one gap seems worth calling out on its own:

**SECURITY.md lists `--enable-script` under "Risky Flags" with "do not enable them unless you have a specific need", but `script` is now part of `basic`, and `basic` is `DEFAULT_PRESET`.** So someone who reads SECURITY.md and turns on no risky flags is still running with `evaluate_script` available, while the page tells them the opposite.

Two smaller things in the same area:

- `--unrestricted-save-paths` is not in "Risky Flags" at all, even though it lifts both save-path restrictions in `assertAllowedPath`.
- SECURITY.md points at `--set-pref`, which is not a real flag. The flag is `--pref name=value`.

## What I changed

Documentation only, no code.

**README.md**
- Corrected the preset table, the module list (`downloads` was missing) and the `--tool-preset developer` example.
- Added `--unrestricted-save-paths` to the CLI options list.
- Split the prefs tools out of the "Firefox Management" line, since `prefs` is a privileged module.
- Replaced "silently skips" with a note that `buildToolset` logs a warning naming the dropped modules.

**SECURITY.md**
- Added a "Tool modules and presets" section that describes capabilities per preset, since that is how the server is configured now. The deprecated flag names are still mentioned so people searching for them land in the right place.
- Documented that `evaluate_script` is in the default preset, with a pointer to `--tool-preset slim` for anyone who does not want it.
- Added `--unrestricted-save-paths` to "Risky Flags".
- Fixed `--set-pref` to `--pref name=value`.

## How I verified it

I printed the real preset composition from `src/tools/index.ts` and compared it line by line against the tables I wrote:

```
DEFAULT_PRESET = basic
slim       + pages, snapshot, input, screenshot
basic      + downloads, script, utilities, management, webextension, screencast
developer  + debugging, network, console, profiler
mozilla    + prefs, privileged
```

## What I deliberately left alone

Whether `script` belongs in `basic` is a product decision rather than a docs one, so I only wrote down the current behaviour. If you would rather move it back to `developer`, I am happy to do that instead and drop the matching parts of this PR.

I would also be glad to add a small test asserting the README preset list against `PRESETS`, along the lines of `tests/tools/registry.test.ts`, so this cannot drift again. Let me know if you want it here or as a follow-up.

Thanks for the reviews on my earlier patches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Browser MCP servers carry inherent risks. A few key practices:
 
 - **Use a dedicated Firefox profile.** Never run the server against your regular profile — the agent has access to whatever the browser can reach, including cookies and saved sessions.
 - **Be cautious about which sites you visit.** Pages can return content designed to manipulate the agent (prompt injection). Stick to sites you control or trust.
-- **Enable only the tool modules you need.** Higher presets such as `--tool-preset developer` (script, debugging) and `--tool-preset mozilla` (privileged context) significantly expand what the agent can do.
+- **Enable only the tool modules you need.** The default `basic` preset already includes `evaluate_script`; `--tool-preset slim` drops it. Higher presets such as `--tool-preset developer` (debugging, network, console, profiler) and `--tool-preset mozilla` (privileged context) expand what the agent can do further.
 
 See [SECURITY.md](SECURITY.md) for a full breakdown of risks and how to report vulnerabilities.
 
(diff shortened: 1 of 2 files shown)


juliandescottes (approved):
Thanks for the cleanup @freema ! 

Especially the preset description which was completely wrong and confusing.

PR #153, opened 2026-08-16
feat: add type_text tool

freema:
Adds `type_text`, a companion to `press_key` for elements that only react to real key-by-key typing (autocomplete fields, rich text editors).

- `text` (required), `uid` (optional, focused first like `press_key`), `submitKey` (optional, same syntax as `press_key`)
- Text and submit key run in one action sequence, so the key lands on the same element
- Reuses `parseKeyCombo` and the focus handling from #82; the key press sequencing is shared with `press_key`
- Unit tests for the handler and the action sequence, integration tests against a real Firefox, `docs/tools.md` regenerated

### DIFF
--- a/tests/integration/type-text.integration.test.ts
+++ b/tests/integration/type-text.integration.test.ts
@@ -16,7 +16,7 @@ describe('type_text Integration Tests', () => {
 
   beforeAll(async () => {
     firefox = await createTestFirefox();
-  }, 30000);
+  });
 
   afterAll(async () => {
     await closeFirefox(firefox);
--- a/tests/integration/type-text.integration.test.ts
+++ b/tests/integration/type-text.integration.test.ts
@@ -1,0 +1,98 @@
+/**
+ * Integration tests for type_text
+ * Tests with real Firefox browser in headless mode
+ */
+
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { createTestFirefox, closeFirefox, waitForElementInSnapshot } from '../helpers/firefox.js';
+import type { FirefoxClient } from '@/firefox/index.js';
+
+const PAGE =
+  'data:text/html,<h1 id=h>Title</h1><input id=a><input id=b>' +
+  '<form id=f onsubmit="document.title=\'submitted\';return false"><input id=c></form>';
+
+describe('type_text Integration Tests', () => {
+  let firefox: FirefoxClient;
+
+  beforeAll(async () => {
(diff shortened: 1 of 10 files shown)


juliandescottes:
@freema thanks for the PR, I will take a look, but did you check #82 by any chance. The timing is unfortunate but both PRs seem to overlap - and #82 was just updated as well.

I will compare them locally, but you might have an easier time since you authored this one.

juliandescottes:
@freema I would try to finalize and merge #82 first and then rebase your work on top of it.

freema:
Hi @juliandescottes, thanks for taking a look! And sorry — that's a fail on my side, I should have checked the open PRs before starting this one and I completely missed #82.

I fully agree with your suggestion: let's finalize and merge #82 first, and once it lands I'll rebase my branch on top of it, so my PR will effectively shrink to `type_text` plus whatever is still useful on top.

For a bit of context: I opened this PR because I need keyboard input for e2e testing of the login flows on my magazine websites, so I'm happy to help with reviewing/testing #82 to move things along.

juliandescottes (changes requested):
#82 is now merged, can you rebase your work on main @freema ? Thanks!

freema:
Rebased on main now that #82 is in. The PR is down to type_text, built on the press_key parser and focus handling: text, optional uid, optional submitKey pressed in the same action sequence. Docs regenerated.

tests/integration/type-text.integration.test.ts
```diff
@@ -0,0 +1,98 @@
...
+const PAGE =
+  'data:text/html,<h1 id=h>Title</h1><input id=a><input id=b>' +
+  '<form id=f onsubmit="document.title=\'submitted\';return false"><input id=c></form>';
+
+describe('type_text Integration Tests', () => {
+  let firefox: FirefoxClient;
+
+  beforeAll(async () => {
+    firefox = await createTestFirefox();
+  }, 30000);
```
juliandescottes: I've been removing those recently to avoid relatively frequent timeouts. The default is now set to 60s. 
```suggestion
  });
```
see https://github.com/mozilla/firefox-devtools-mcp/commit/df056329cbfdd8cd83904d64e7622d17379e05b3

juliandescottes (approved):
Looks good, just one hardcoded 30s timeout to remove per a recent change. 
If you have time to apply it that's great, otherwise I'll merge and follow up!

PR #154, opened 2026-08-17
Bug 2063831 - Require --android-wipe-app-data when connecting to Android

juliandescottes:
Until we can safely connect without wiping user data, update documentation and require an explicit flag to be passed when connecting to Android.

### DIFF
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -166,9 +166,25 @@ export class FirefoxCore {
    */
   async connect(): Promise<void> {
     const isAndroid = this.options.androidDevice !== undefined;
+    const androidPackage = this.options.androidPackage ?? 'org.mozilla.firefox';
+
+    if (isAndroid && !this.options.androidWipeAppData) {
+      // geckodriver runs "adb shell pm clear <package>" before every Android session
+      // (AndroidHandler::prepare) and offers no way to opt out, so launching wipes the
+      // data of the target app instead of only using its own temporary profile.
+      // Bug 2064088 tracks adding an opt-out to geckodriver.
+      throw new Error(
+        `Firefox for Android mode wipes all data of ${androidPackage} ` +
+          'on the device: tabs, history, bookmarks, passwords, cookies and settings are all lost, ' +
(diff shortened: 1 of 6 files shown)


hbenl (approved):

juliandescottes:
Thanks for the review!

src/firefox/core.ts
```diff
@@ -166,9 +166,25 @@ export class FirefoxCore {
...
+    const androidPackage = this.options.androidPackage ?? 'org.mozilla.firefox';
+
+    if (isAndroid && !this.options.androidWipeAppData) {
+      // geckodriver runs "adb shell pm clear <package>" before every Android session
+      // (AndroidHandler::prepare) and offers no way to opt out, so launching wipes the
+      // data of the target app instead of only using its own temporary profile.
+      // Bug 2064088 tracks adding an opt-out to geckodriver.
+      throw new Error(
+        `Firefox for Android mode wipes all data of ${androidPackage} ` +
+          'on the device: tabs, history, bookmarks, passwords, cookies and settings are all lost, ' +
```
whimboo, line 178: @juliandescottes this isn't correct here. Tabs, history, bookmarks and others are part of the Firefox profile, while only certain app settings including permissions are getting removed. See https://bugzilla.mozilla.org/show_bug.cgi?id=2064088#c4 for details.

juliandescottes: @whimboo Your description does not match what users reported:

> All app user data is destroyed. Onboarding re-runs on next launch; history, open tabs, saved logins, bookmarks, extensions and settings are gone. No warning, no confirmation prompt, no mention in any documentation I read before running it.

whimboo: Do we duplicate all that data for app settings beside the usual profile usage? That's strange.

juliandescottes: See https://bugzilla.mozilla.org/show_bug.cgi?id=2064088#c11

1/ The profile doesn't hold as much information on desktop compared to Android
2/ The real (non-automation) profile is stored in app-data as well and gets fully erased by pm clear.

So the warning is correct.

PR #155, opened 2026-08-18
chore: prepare release v0.10.0

juliandescottes:


juliandescottes:
`--android-wipe-app-data` is a breaking change and this is the first release where the MCPB should be able to pickup the connection port on its own.

hbenl (approved):

juliandescottes:
Thanks! Will try to do a release, hopefully I get it right the first time here

PR #156, opened 2026-08-19
refactor: add BiDiFacade and use it in event handling modules

hbenl:


src/firefox/bidi.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  async sendCommand(method: string, params: Record<string, any> = {}): Promise<any> {
+    const bidi = await this.driver.getBidi();
+    // bidi.socket is a Node.js `ws` WebSocket (EventEmitter-style), but typed as browser WebSocket
+    const ws = bidi.socket as any;
+
+    // Wait for WebSocket to be ready before sending
+    await this.waitForWebSocketOpen(ws);
+
+    const id = Math.floor(Math.random() * 1000000);
```
juliandescottes: followup: this was just moved over, but a random id is not amazing when reading logs (at least for a human).

Counter or uuid would be easier to track.

src/firefox/events/network.ts
```diff
@@ -50,122 +47,113 @@ export class NetworkEvents {
...
+      'network.beforeRequestSent',
+      'network.responseStarted',
+      'network.responseCompleted',
+    ]);
 
     // Subscribe to navigation events for lifecycle hooks
     try {
-      await bidi.subscribe('browsingContext.load', contextId ? [contextId] : undefined);
-      await bidi.subscribe('browsingContext.domContentLoaded', contextId ? [contextId] : undefined);
+      await this.bidi.subscribe(['browsingContext.load', 'browsingContext.domContentLoaded']);
```
juliandescottes: We can clean this up in a follow up, but I don't think the try/catch around the subscribe for `'browsingContext.load', 'browsingContext.domContentLoaded'` makes a lot of sense if the other events are not in a try / catch? 

load/domContentLoaded should be available anywhere BiDi is available for Firefox.

Other modules seem to use the same pattern.

src/firefox/bidi.ts
```diff
@@ -0,0 +1,105 @@
...
+      }, 10000);
+    });
+  }
+
+  private listenForEvents(ws: any) {
+    ws.on('message', (data: any) => {
+      try {
+        const payload = JSON.parse(data.toString());
+        if (payload.type === 'event' && payload.method) {
+          this.emit(payload.method, payload.params);
```
juliandescottes: node's `emit` stops on the first listener that throws, so in the future if we have different listeners for the same event, it could be an issue. We can handle that in a follow up.

In any case the `emit` should not be in the try / catch which only expects to swallow parse errors, we should at least log errors triggered by emit.

src/firefox/index.ts
```diff
@@ -39,36 +48,47 @@ export class FirefoxClient {
...
-
-      this.networkEvents = new NetworkEvents(
-        driver as any,
-        {
-          autoClearOnNavigate: false,
-          captureBodies: this.core.getOptions().captureNetworkBodies !== false,
-        },
-        (method, params) => this.core.sendBiDiCommand(method, params ?? {})
-      );
+    // Cast to any for BiDi-specific APIs that only exist on selenium WebDriver
```
juliandescottes: I think the cast as `any` is gone now?

juliandescottes (approved):
Thanks for moving all the bidi related code to a single helper! 

Mostly follow up suggestions on my side.

PR #157, opened 2026-08-19
Bug 2064762 - [firefox-devtools-mcp] Advertise server instructions in the initialize response

juliandescottes:
Claude desktop seems to ignore the MCP and default much more strongly to chrome since a recent update.


juliandescottes:
I wish I could say this is guaranteed to fix the issue, but I can no longer reproduce the original problem. Non-deterministic workflows for the win...

hbenl (approved):

PR #160, opened 2026-08-20
Bug 2065207 - [firefox-devtools-mcp] Update Firefox devtools mcp to use --android-keep-app-data when available

juliandescottes:


juliandescottes:
Moving to draft as we are not sure if `--android-keep-app-data` will land as is.

juliandescottes:
Closing the PR for now, it looks increasingly like this argument will not be added. Will re-open in case anything changes.

PR #161, opened 2026-08-20
test: make profile path assertions cross-platform

f3tchcodes:
## Summary

Make Firefox profile path test assertions platform-independent by using
`node:path.join()` instead of hard-coded path separators.

The production code already uses `path.join()`, so on Windows it correctly
produces Windows-style paths. The tests previously expected `/` separators,
causing them to fail on Windows.

## Testing

- `npx vitest run tests/firefox/profile.test.ts`
- `npx vitest run tests/firefox/core.test.ts`
- `npm run typecheck:tests`

### DIFF
--- a/tests/firefox/core.test.ts
+++ b/tests/firefox/core.test.ts
@@ -2,6 +2,8 @@
  * Unit tests for FirefoxCore module
  */
 
+import { join } from 'node:path';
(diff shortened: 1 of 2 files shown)


juliandescottes (approved):
Thanks a lot for fixing the tests on Windows!

PR #164, opened 2026-08-22
chore: enforce LF line endings with gitattributes

f3tchcodes:
Enforcing LF line endings with gitattributes to avoid linting errors on other platforms (ie. windows).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


juliandescottes (changes requested):
Thanks, could you add the newline at the end of the file? 

I'm not sure if any of the crlf extensions added in the #169 are necessary at this point, but I imagine we can merge your PR first anyway.

f3tchcodes:
Added, thanks!

juliandescottes (approved):
Great, thanks!

PR #165, opened 2026-08-22
test: make process cleanup work on Windows

f3tchcodes:
## Summary

Make the Vitest cleanup work properly on Windows.

The current cleanup uses Unix-only commands like `pgrep`, `pkill`, and `kill`, so running the tests on Windows prints command errors even though the tests themselves still run.

This change keeps the existing cleanup behavior for Linux and macOS, while adding a Windows-specific cleanup path. On Windows, it finds Firefox test instances using the `--marionette` command-line argument, terminates those process trees with `taskkill`, and cleans up leftover `geckodriver.exe` processes without touching normal Firefox windows.

### DIFF
--- a/tests/setup.ts
+++ b/tests/setup.ts
@@ -83,7 +83,7 @@ function findWindowsProcessIds(imageName: string, commandLinePattern?: string): 
   const output = execFileSync(
     'powershell.exe',
     ['-NoProfile', '-NonInteractive', '-Command', script],
-    { encoding: 'utf-8', stdio: ['ignore', 'pipe', 'ignore'] }
+    { encoding: 'utf-8', stdio: ['ignore', 'pipe', 'ignore'], windowsHide: true }
   );
   return output
     .split(/\r?\n/)
--- a/tests/setup.ts
+++ b/tests/setup.ts
@@ -71,8 +71,10 @@ function cleanupUnix() {
   }
 }
 
-// Only matches Firefox instances whose command line carries --marionette,
-// so regular user-launched Firefox windows are never touched.
+/**
+ * Windows only. Retrieve all process ids matching the provided name and
+ * optional command line pattern.
+ */
 function findWindowsProcessIds(imageName: string, commandLinePattern?: string): number[] {
   const filter = commandLinePattern
     ? `Name='${imageName}' AND CommandLine LIKE '%${commandLinePattern}%'`
--- a/tests/setup.ts
+++ b/tests/setup.ts
@@ -2,7 +2,9 @@
 // This file runs before all tests
 
 import { beforeAll, afterAll } from 'vitest';
-import { execSync } from 'child_process';
+import { execSync, execFileSync } from 'child_process';
+
+const isWindows = process.platform === 'win32';
 
 // Track if we're in cleanup mode
 let isCleaningUp = false;
@@ -17,7 +19,7 @@ afterAll(() => {
 });
 
 /**
- * Cleanup function to kill all Firefox and geckodriver processes
+ * Cleanup function to kill leftover test Firefox and geckodriver processes
  * This ensures no zombie processes are left after test runs
  */
 function cleanup() {
@@ -26,6 +28,20 @@ function cleanup() {
   }
   isCleaningUp = true;
 
+  try {
+    if (isWindows) {
+      cleanupWindows();
+    } else {
+      cleanupUnix();
+    }
+  } catch {
+    // Ignore errors - processes might already be dead
+  } finally {
+    isCleaningUp = false;
+  }
+}
+
+function cleanupUnix() {
   try {
     // Find Firefox processes started with --marionette (test instances)
     const firefoxPids = execSync('pgrep -f "firefox.*marionette" || true', {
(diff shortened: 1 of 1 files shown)


tests/setup.ts
```diff
@@ -26,6 +28,20 @@ function cleanup() {
...
+    if (isWindows) {
+      cleanupWindows();
+    } else {
+      cleanupUnix();
+    }
+  } catch {
+    // Ignore errors - processes might already be dead
+  } finally {
+    isCleaningUp = false;
+  }
```
juliandescottes: Both `cleanupWindows` and `cleanupUnix` are already wrapped in a try/catch, so this one should be unnecessary?

```suggestion
  if (isWindows) {
    cleanupWindows();
  } else {
    cleanupUnix();
  }
  isCleaningUp = false;
```

(or remove the try/catch from the individual functions?)

tests/setup.ts
```diff
@@ -57,8 +73,55 @@ function cleanup() {
...
+    .map(Number);
+}
+
+function cleanupWindows() {
+  try {
+    const firefoxPids = findWindowsProcessIds('firefox.exe', '--marionette');
+    for (const pid of firefoxPids) {
+      try {
+        // /T kills the whole process tree, replacing the Unix child-kill step
+        execSync(`taskkill /PID ${pid} /T /F`, { stdio: 'ignore' });
```
juliandescottes, line 103: Looking at https://github.com/mozilla/firefox-devtools-mcp/pull/169/ , what about the `windowsHide` argument to prevent spawning command windows?

tests/setup.ts
```diff
@@ -57,8 +73,55 @@ function cleanup() {
...
   } catch {
     // Ignore errors - processes might already be dead
-  } finally {
-    isCleaningUp = false;
+  }
+}
+
+// Only matches Firefox instances whose command line carries --marionette,
+// so regular user-launched Firefox windows are never touched.
+function findWindowsProcessIds(imageName: string, commandLinePattern?: string): number[] {
```
juliandescottes: The command mentions it finds "firefox" instances with the `--marionette` cli arg, so it's odd to have the "imageName" and "commandLinePattern" arguments. 

Let's simplify this and remove the arguments.

juliandescottes (changes requested):
Thanks! 

It looks compatible with the changes from https://github.com/mozilla/firefox-devtools-mcp/pull/169 . I hope @freema doesn't mind, I think it will be easier to land your PR first and then trim down 169 to be a bit smaller.

A few comments to address first.

tests/setup.ts
juliandescottes: Oh I missed that this was also used for geckodriver. Let's just fix the comment to make it match the behavior.

f3tchcodes:
All changes have been made! I also renamed the function `findWindowsProcessIds()` to `findMarionetteFirefoxProcessIds()` for better readability.

tests/setup.ts
```diff
@@ -57,8 +68,54 @@ function cleanup() {
...
+    for (const pid of firefoxPids) {
+      try {
+        // /T kills the whole process tree, replacing the Unix child-kill step
+        execSync(`taskkill /PID ${pid} /T /F`, { stdio: 'ignore' });
+      } catch {
+        // Ignore errors - process might already be dead
+      }
+    }
+
+    for (const pid of findMarionetteFirefoxProcessIds()) {
```
juliandescottes: This used to find geckodriver processes, but this got lost because of my comment. Can you restore the previous version?

tests/setup.ts
juliandescottes: You resolved the conversation, can you confirm if `windowsHide` is not needed here?

juliandescottes (changes requested):
My bad, the `findWindowsProcessIds` was also used to find geckodriver processes, only the comment was confusing. We need to restore it and fix the comment

tests/setup.ts
f3tchcodes: Yes, `windowsHide` is useful here to prevent the PowerShell process from briefly spawning a visible window. I added it in the latest changes - sorry, I resolved the conversation before confirming.

f3tchcodes:
Restored! I completely missed it even after reading through the code haha. Thanks for catching it before the merge!

tests/setup.ts
```diff
@@ -57,8 +68,55 @@ function cleanup() {
     console.log('✅ Global cleanup: All test Firefox processes terminated');
   } catch {
     // Ignore errors - processes might already be dead
-  } finally {
-    isCleaningUp = false;
+  }
+}
+
+// Only matches Firefox instances whose command line carries --marionette,
+// so regular user-launched Firefox windows are never touched.
```
juliandescottes: We should still update this comment, this is a generic helper it's not dedicated to finding Firefox instances with --marionette.

(this is what confused me in my first review)
```suggestion
/**
 * Windows only. Retrieve all process ids matching the provided name and 
 * optional command line pattern.
 */
```

Also why imageName ? Does that make sense in the windows domain ?

juliandescottes (changes requested):
Just a comment to update and we should be good to go, thanks!

tests/setup.ts
f3tchcodes: `imageName` comes from Windows terminology, apparently tools like `tasklist` call the executable column **"Image Name"**, and `taskkill /IM` means "image name", so it's technically not wrong.

But if you'd still prefer something clearer like `processName`, let me know and I can change it.

juliandescottes (approved):
Thanks!

PR #166, opened 2026-08-22
chore: improve tool input schema typing

f3tchcodes:
Improving tools defination typing by adding interface for inputSchema.

### DIFF
--- a/src/tools/console.ts
+++ b/src/tools/console.ts
@@ -12,7 +12,7 @@ import {
   truncationFooter,
 } from '../utils/response-helpers.js';
 import { saveOutput } from '../utils/save-output.js';
-import { defineModule } from './module.js';
+import { defineModule, type ToolDefinition } from './module.js';
 import type { McpToolResponse } from '../types/common.js';
 
 export const listConsoleMessagesTool = {
@@ -63,7 +63,7 @@ export const listConsoleMessagesTool = {
       },
     },
   },
-};
+} satisfies ToolDefinition;
 
 export const clearConsoleMessagesTool = {
   name: 'clear_console_messages',
@@ -75,7 +75,7 @@ export const clearConsoleMessagesTool = {
     type: 'object',
     properties: {},
   },
-};
+} satisfies ToolDefinition;
 
 const DEFAULT_LIMIT = 50;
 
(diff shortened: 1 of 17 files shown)


juliandescottes (changes requested):
Before we merge this, what is the intent: documentation (+ code hints etc...) or strict typechecks? 

If we want to enforce a strict typecheck, at the moment it's not enough because many modules define tools which have very simple schemas. Example for clearConsole:
```
  inputSchema: {
    type: 'object',
    properties: {},
  },
```

Because of this above, all the other Console tools' `properties` will not be strictly checked (it's apparently known as subtype reduction, but I don't know much about typescript). And as I said that's the case for most modules at the moment.

If we want to enforce it, we would need to annotate all tools with `satisfies ToolDefinition`.

While I am a bit worried about freezing the shape of schema based on the current tools, at least if we have consistent failures when a tool no longer matches it, it will be easy to maintain.

Can you annotate the tools as suggested?

f3tchcodes:
Yes, the intent was compile time checking.

The reason I initially didn't add `satisfies ToolDefinition` to every tool definition was mainly to avoid repeating it across every tool, since all of the definitions get passed through `defineModule(...)` for typechecking at the end anyway.

I expected the `ToolDefinition` typing there to provide the compile time checking at `defineModule(...)`, but after some more research I see what you mean about subtype reduction allowing some definitions to avoid being checked as strictly as intended.

I can add `satisfies ToolDefinition` to each tool definition so the check is enforced, let me know if that's fine.

That should also let us make the schema typing stricter (adding stricter types for `type` property rather than keeping it `string`), since each tool defination would now be checked against `ToolDefinition` at declaration.

juliandescottes:
> I expected the `ToolDefinition` typing there to provide the compile time checking at `defineModule(...)`, [...]
> 

Same here, I was quite surprised it didn't work out of the box.

> I can add `satisfies ToolDefinition` to each tool definition so the check is enforced, let me know if that's fine.
> 

Yes, let's add the annotation, thanks!

juliandescottes (changes requested):
Let's add the annotations :)

f3tchcodes:
Added!

juliandescottes (approved):
Thanks!

PR #168, opened 2026-08-22
Bug 2065545 - [firefox-devtools-mcp] Add a tool to disable the network cache

shoemoney:
Adds a `set_network_cache` tool so an agent can send every request to the network instead of the cache:

```js
{ "behavior": "bypass" }                    // selected tab
{ "behavior": "bypass", "scope": "global" } // browser-wide
{ "behavior": "default" }                   // restore
```

Per your note on the bug, this is a tool rather than a config, and it is scoped to the selected tab by default with `scope: "global"` as the opt-in. It joins the existing `network` module, so it arrives with the `developer` preset alongside the profiler tools it is meant to be used with.

### One correction to the bug, with evidence

The bug says:

> WebDriver BiDi supports the `emulation.setNetworkConditions` command, which allows to bypass the cache.

That command does exist in Firefox, but it has no cache parameter: it is offline emulation. Probed against Firefox 154 through this repo's own BiDi transport:

```
ERR   emulation.setNetworkConditions {contexts, cacheDisabled:true}  -> invalid argument
ERR   emulation.setNetworkConditions {contexts, offline:false}       -> invalid argument
ERR   emulation.setNetworkConditions {contexts, latency:0}           -> invalid argument
ERR   emulation.setNetworkConditions networkConditions:"offline"     -> invalid argument
OK    emulation.setNetworkConditions networkConditions:{type:"offline"}
OK    emulation.setNetworkConditions networkConditions:null
```

The only accepted values are `{type: "offline"}` and `null`.

I checked that `invalid argument` really meant "exists, wrong params" rather than "no such command", because reading it the other way would have sent me down the wrong path:

```
ERR   bogus.noSuchCommand                    -> unknown command
ERR   emulation.noSuchThing                  -> unknown command
ERR   emulation.setLocaleOverride {}         -> invalid argument   (real command, bad params)
```

The command that actually bypasses the cache is `network.setCacheBehavior`, with `cacheBehavior: "bypass" | "default"` and an optional `contexts`. Omit it for global; pass one for per-tab. That is what this PR wraps.

### Measured, not assumed

A local server serving `asset.js` with `Cache-Control: public, max-age=86400`, counting real server hits across 3 page loads:

| state | 3 loads → server hits |
|---|---|
| baseline (default) | **1** (cached) |
| `setCacheBehavior('bypass')`, tab-scoped | **3** (every load hit the network) |
| `setCacheBehavior('default')` restore | 0 (served from cache again) |
| `setCacheBehavior('bypass', {global:true})` | **3** |

Both scopes work, and the effect is reversible.

### Rejecting rather than defaulting

An unrecognised `behavior` or `scope` returns an error instead of falling back:

```
behavior must be one of default, bypass (got "disabled")
scope must be 'tab' or 'global' (got "everything")
```

Defaulting to `default` when someone asked for `disabled` would silently leave the cache on and quietly invalidate whatever they were measuring. This is the exact failure this tool exists to prevent. `bypass` also echoes back how to undo it, since the setting persists until reset:

```
cache bypass (selected tab): set behavior='default' to restore caching
```

### Tests

Six new unit tests in `tests/tools/network.test.ts`, all failing on unmodified `main`:

```
× exposes a behavior enum and an optional scope
× scopes to the selected tab by default
× applies browser-wide when scope is global
× tells the caller how to restore caching after a bypass
× rejects an unknown behavior instead of defaulting to cached
× rejects an unknown scope instead of silently using the tab
```

No integration test: the cache-bypass assertion needs a local HTTP server with a cacheable asset and a hit counter, which is a new pattern for `tests/integration/` (every file there uses `file://` fixtures, and `file://` cannot exercise the HTTP cache). I can add one if you would like that pattern introduced. The measurements above came from exactly that harness, so it is written.

```
npm run lint            clean
npm run format:check    clean
npm run typecheck       clean
npm run typecheck:tests clean
npm run build           clean
npm run test:unit       43 files, 601 tests passed
```

### Left alone deliberately

`emulation.setNetworkConditions` with `{type: "offline"}` is a genuinely useful thing to expose for offline-mode testing, but it is a different feature from this bug and would want its own tool and its own name. I can follow up if you want it.

### DIFF
--- a/src/firefox/index.ts
+++ b/src/firefox/index.ts
@@ -12,12 +12,8 @@ import { ConsoleEvents, NetworkEvents, DebuggingEvents, DownloadEvents } from '.
 import type { NetworkBodyResult } from './events/network.js';
 import { DomInteractions } from './dom.js';
 import { PageManagement, type ReadinessState } from './pages.js';
-import { CacheManagement, CACHE_BEHAVIORS, isCacheBehavior, type CacheBehavior } from './cache.js';
+import { CacheManagement, type CacheBehavior } from './cache.js';
 import { SnapshotManager, type Snapshot, type SnapshotOptions } from './snapshot/index.js';
-
-// Re-exported for backward compatibility: src/tools/network.ts imports
-// these cache types from firefox/index.js rather than firefox/cache.js.
-export { CACHE_BEHAVIORS, isCacheBehavior, type CacheBehavior };
 
 /**
  * Main Firefox Client facade
--- a/src/firefox/index.ts
+++ b/src/firefox/index.ts
@@ -12,20 +12,12 @@ import { ConsoleEvents, NetworkEvents, DebuggingEvents, DownloadEvents } from '.
 import type { NetworkBodyResult } from './events/network.js';
 import { DomInteractions } from './dom.js';
 import { PageManagement, type ReadinessState } from './pages.js';
+import { CacheManagement, CACHE_BEHAVIORS, isCacheBehavior, type CacheBehavior } from './cache.js';
 import { SnapshotManager, type Snapshot, type SnapshotOptions } from './snapshot/index.js';
 
-/**
- * WebDriver BiDi network.CacheBehavior.
- * - "default": normal HTTP cache behaviour
- * - "bypass": skip the cache, so every request goes to the network
- */
-export const CACHE_BEHAVIORS = ['default', 'bypass'] as const;
-
-export type CacheBehavior = (typeof CACHE_BEHAVIORS)[number];
-
-export function isCacheBehavior(value: unknown): value is CacheBehavior {
-  return CACHE_BEHAVIORS.includes(value as CacheBehavior);
-}
+// Re-exported for backward compatibility: src/tools/network.ts imports
+// these cache types from firefox/index.js rather than firefox/cache.js.
+export { CACHE_BEHAVIORS, isCacheBehavior, type CacheBehavior };
 
 /**
  * Main Firefox Client facade
@@ -40,6 +32,7 @@ export class FirefoxClient {
   private downloadEvents: DownloadEvents | null = null;
   private dom: DomInteractions | null = null;
   private pages: PageManagement | null = null;
+  private cache: CacheManagement | null = null;
   private snapshot: SnapshotManager | null = null;
 
   constructor(options: FirefoxLaunchOptions) {
@@ -112,6 +105,11 @@ export class FirefoxClient {
       driver,
       () => this.core.getCurrentContextId(),
       (id: string) => this.core.setCurrentContextId(id),
+      (method: string, params: Record<string, any>) => this.getBidi().sendCommand(method, params)
+    );
+
+    this.cache = new CacheManagement(
+      () => this.core.getCurrentContextId(),
       (method: string, params: Record<string, any>) => this.getBidi().sendCommand(method, params)
     );
   }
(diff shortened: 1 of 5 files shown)


src/firefox/index.ts
```diff
@@ -14,6 +14,19 @@ import { DomInteractions } from './dom.js';
...
+ * - "default": normal HTTP cache behaviour
+ * - "bypass": skip the cache, so every request goes to the network
+ */
+export const CACHE_BEHAVIORS = ['default', 'bypass'] as const;
+
+export type CacheBehavior = (typeof CACHE_BEHAVIORS)[number];
+
+export function isCacheBehavior(value: unknown): value is CacheBehavior {
+  return CACHE_BEHAVIORS.includes(value as CacheBehavior);
+}
```
juliandescottes: Feels a bit odd in index.ts. Could we move this to another module, eg create a `src/firefox/cache.ts` ?

juliandescottes (changes requested):
Thanks for the patch, works nicely. Just one small comment to address.

shoemoney:
Done, moved the cache pieces to src/firefox/cache.ts in 4923641.

src/firefox/index.ts
```diff
@@ -12,8 +12,13 @@ import { ConsoleEvents, NetworkEvents, DebuggingEvents, DownloadEvents } from '.
 import type { NetworkBodyResult } from './events/network.js';
 import { DomInteractions } from './dom.js';
 import { PageManagement, type ReadinessState } from './pages.js';
+import { CacheManagement, CACHE_BEHAVIORS, isCacheBehavior, type CacheBehavior } from './cache.js';
 import { SnapshotManager, type Snapshot, type SnapshotOptions } from './snapshot/index.js';
 
+// Re-exported for backward compatibility: src/tools/network.ts imports
+// these cache types from firefox/index.js rather than firefox/cache.js.
+export { CACHE_BEHAVIORS, isCacheBehavior, type CacheBehavior };
```
juliandescottes: Not sure about the backward compatibility issue?

`src/tools` and `src/firefox` should always be in sync, so `src/tools/network.ts` can import from `cache.ts` directly?

src/tools/network.ts
```diff
@@ -131,6 +132,31 @@ export const getNetworkRequestTool = {
   },
 } satisfies ToolDefinition;
 
+export const setNetworkCacheTool = {
+  name: 'set_network_cache',
+  description:
+    "Control the HTTP cache. Use behavior='bypass' so every request goes to the network — useful for performance measurement and for verifying a change that a cached asset would otherwise hide. Applies to the selected tab unless scope='global'. Persists until set back to 'default'.",
```
juliandescottes:
```suggestion
    "Control the HTTP cache. Use behavior='bypass' so every request goes to the network — useful for performance measurement and for verifying a change that a cached asset would otherwise hide. Applies to the currently selected tab unless scope='global'. Persists until set back to 'default' or Firefox shuts down.",
```

Minor modifications for the description. Should make it clear that the behavior will not "move" to another tab when it gets selected. Also the MCP can restart Firefox and the behavior would be lost at this point.

src/tools/network.ts
```diff
@@ -576,11 +602,43 @@ export const handleGetNetworkRequest = defineToolHandler(
   }
 );
 
+export async function handleSetNetworkCache(args: unknown): Promise<McpToolResponse> {
```
juliandescottes: Note: I can update it in a follow up but since PR #171, we can use defineToolHandler to avoid the duplicated try / catch + error handling

juliandescottes (changes requested):
Thanks for the update, I missed some items in my first review, sorry about that !

shoemoney:
Addressed all three:
- Dropped the backward-compat re-export in `firefox/index.ts`; `tools/network.ts` now imports `CACHE_BEHAVIORS`/`isCacheBehavior` straight from `firefox/cache.js`.
- Applied your suggested description wording verbatim.
- Switched `handleSetNetworkCache` to `defineToolHandler`, matching `handleGetNetworkRequest`.

Also rebased onto main to pick up #171. Lint/format/typecheck/build/tests all green (655 tests, incl. the doc-sync check after the description edit).

juliandescottes (approved):
Thanks @shoemoney, that looks good to me. We can follow up and update the performance skills now !

Issue #33, opened 2026-01-30
Windows: Vitest fork issues with Firefox integration tests


freema:
## Problem

Vitest has known issues with process forking on Windows when running integration tests that spawn Firefox. The tests hang or fail unpredictably due to how vitest isolates test processes.

Related upstream issues:
- https://github.com/elastic/kibana/issues/52053
- https://github.com/vitest-dev/vitest/issues/317

## Current Workaround

We use a separate test runner (`scripts/run-integration-tests-windows.mjs`) that runs integration tests directly via Node.js without vitest's process isolation.

The CI workflow detects Windows and automatically uses this runner instead of vitest for integration tests. Unit tests still run via vitest on all platforms.

## Ideal Solution

Once vitest fixes the underlying forking issues on Windows, we should:
1. Remove the separate Windows test runner
2. Use vitest for all tests on all platforms

## Labels

- platform: windows
- type: workaround

PR #169, opened 2026-08-23
test(windows): run the integration suite on Windows

freema:
Trimmed to the one commit left after the split: #172 (paths), #173 (dev tooling) and #174 (Firefox binary detection) carry the rest, and the `tests/setup.ts` cleanup now comes from #165.

What it does:
- drops the Windows exclusion in `vitest.config.ts` and the standalone runner (`scripts/run-integration-tests-windows.mjs`, `test:integration:win`); the vitest fork hang from #33 no longer reproduces on Vitest 4
- builds fixture URLs with `pathToFileURL` (`fixtureUrl()`), packs the extension fixture in-process instead of shelling out to `zip`, and retries temp-dir removal (`removeDir()`)
- runs `ci.yml` on `windows-latest` next to `ubuntu-latest`

Verified on Windows (7 integration files, 54 tests) and macOS (unit 608, integration 54).


f3tchcodes:
FYI, there's some overlap with #165 which already covers Windows process cleanup in `tests/setup.ts,` and #164 covers the `.gitattributes` LF fix. Mentioning in case it's useful to avoid duplicate work while reviewing this PR.

CHANGELOG.md
```diff
@@ -7,6 +7,45 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
```
juliandescottes: This is too much for a changelog. Can you keep it short and focus on the impact for users?

src/firefox/binary.ts
```diff
@@ -0,0 +1,133 @@
+/**
+ * Firefox binary discovery.
+ *
+ * On Windows geckodriver only searches the Program Files directories and
+ * HKEY_LOCAL_MACHINE, so it cannot see a per-user install (%LOCALAPPDATA%,
+ * registered under HKCU) — what the installer produces without admin rights.
+ * Finding the binary here and passing it as moz:firefoxOptions.binary fixes
+ * that. Windows-only: geckodriver's own lookup suffices elsewhere.
```
juliandescottes: This sounds like something which should be fixed upstream in geckodriver rather than handled in the MCP with a workaround.

docs/testing.md
```diff
@@ -39,6 +39,31 @@ All tests are self-contained (no ordering dependencies) and use active polling
...
+- **Deleting temp dirs**: use `removeDir()` from `tests/helpers/fs.ts`. Plain
+  `rmSync({ force: true })` intermittently throws `ENOTEMPTY` while handles close.
+- **Line endings**: `.gitattributes` checks out text files as LF. Without it
+  `core.autocrlf=true` makes `npm run format:check` fail on every file.
+- **Path comparisons**: compare against a root with `isWithinRoot()`
+  (`src/utils/save-output.ts`). Windows paths are case-insensitive, so a plain
+  `startsWith` rejects valid paths that differ only in case.
+- **`process.env` in tests**: the real environment is case-insensitive on Windows
+  (`process.env.SystemRoot` resolves the `SYSTEMROOT` key). Replacing it with a
+  plain object drops that, so set the variables a test needs explicitly.
```
juliandescottes: This mostly repeats what the commit has been fixing. The documentation files are quite big already, I would skip that.

docs/ci-and-release.md
```diff
@@ -37,12 +37,11 @@ Release flow
...
-- Unit tests still run via vitest on all platforms.
+Windows
+- CI runs the full suite — unit *and* integration tests — on `windows-latest` as well as `ubuntu-latest`.
+- vitest used to hang when forking integration tests that spawn Firefox on Windows (issue #33), so those tests were excluded there and a standalone runner (`scripts/run-integration-tests-windows.mjs`) stood in for them. The hang no longer reproduces on vitest 4, so the exclusion and the standalone runner were both removed and Windows runs the same tests as every other platform.
+- geckodriver only searches the Program Files directories and HKEY_LOCAL_MACHINE for Firefox, so it cannot see a per-user install (`%LOCALAPPDATA%\Mozilla Firefox`), which is what the installer produces without administrator rights. `src/firefox/binary.ts` locates the binary itself and passes it as `moz:firefoxOptions.binary`; use `--firefox-path` if an install still is not found.
+- `.gitattributes` checks out text files with LF everywhere. Without it, Git for Windows' `core.autocrlf=true` gives the working tree CRLF endings and `npm run format:check` fails on every file.
```
juliandescottes: Repeats what the commit does, not worth keeping in the docs.

scripts/build-mcpb.mjs
```diff
@@ -35,9 +35,9 @@ try {
   );
 
   const outDir = resolve(root, 'dist-mcpb');
-  if (!existsSync(outDir)) {
-    execSync(`mkdir -p ${outDir}`);
-  }
+  // mkdirSync rather than shelling out to `mkdir -p`, which does not exist in
+  // the cmd.exe that npm runs scripts through on Windows.
```
juliandescottes: I would remove this, it only make sense when looking at the diff.

src/firefox/core.ts
```diff
@@ -308,6 +310,9 @@ export class FirefoxCore {
       const firefoxOptions = new firefox.Options();
       firefoxOptions.enableBidi();
 
+      // True when no binary was found on Windows, so the failure can say why.
+      let binaryLookupFailed = false;
```
juliandescottes: The variable name should make it clear this is windows only

src/firefox/binary.ts
```diff
@@ -0,0 +1,133 @@
...
+  return findFirefoxBinaryWindows();
+}
+
+/** Replaces geckodriver's "unable to find binary" error, which omits the fix. */
+export const WINDOWS_BINARY_NOT_FOUND_HINT =
+  'Firefox could not be found on this system. geckodriver only searches the Program Files ' +
+  'directories and HKEY_LOCAL_MACHINE, so a Firefox installed for the current user only is ' +
+  'not detected automatically. Pass the full path to firefox.exe via --firefox-path ' +
+  '(for example --firefox-path "%LOCALAPPDATA%\\Mozilla Firefox\\firefox.exe"), or reinstall ' +
+  'Firefox for all users.';
```
juliandescottes: I thought this binary.ts helper was meant to bypass the geckodriver limitation, but the message says that user installed Firefox are not handled? 

I would go for a shorter message. eg

> Unable to detect Firefox binary automatically, please provide the full path via --firefox-path

Also, why define this constant here if it's only used in core.ts?

juliandescottes (changes requested):
Thanks for the patch. I haven't fully tested, but it would be great to have more granular PRs. Could you split this ?

.gitattributes
```diff
@@ -0,0 +1,9 @@
+# Check out text files with LF everywhere. Without this, Git for Windows'
+# core.autocrlf=true gives the working tree CRLF endings, which collide with
+# Prettier's `endOfLine: "lf"` and make `npm run format:check` fail on every file.
+* text=auto eol=lf
+
+# Windows shell helpers must keep CRLF.
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.ps1 text eol=crlf
```
juliandescottes: I don't see any file with those extensions in the repository, is this necessary? Otherwise this would be covered by #164 already?

src/firefox/binary.ts
juliandescottes: I think this is captured at https://bugzilla.mozilla.org/show_bug.cgi?id=1921933

(Which is a mentored bug, in case you're interested to do a geckodriver contribution :) )

freema:
Trimmed as discussed, this is now only the integration-suite commit on top of #165; the rest lives in #172, #173 and #174, and the docs/changelog additions went with it. Thanks for the split suggestion.

tests/helpers/zip.ts
```diff
@@ -0,0 +1,123 @@
+/**
```
juliandescottes: Is there a strong reason to re-implement a zip writer here? We already pull: 
- fflate
- jszip
- zip.js

fflate seems the most straightforward choice here. It's lightweight and has a sync API. Can you try to use it instead?

tests/tools/console.test.ts
```diff
@@ -100,7 +101,7 @@ describe('Console Tools', () => {
     afterEach(() => {
       vi.restoreAllMocks();
       if (existsSync(tempDir)) {
```
juliandescottes, line 103: Nothing new from your PR, but in theory force: true ignores errors if the path does not exist, so those guards are most likely redundant. We could remove them in another PR.

juliandescottes (changes requested):
Thanks, overall this looks good, but I don't think we should re-implement a zip helper here, there are many options to choose from.

tests/helpers/zip.ts
freema: Replaced with fflate (added as a devDependency, it was only transitive before), the helper file is gone. Full suite passes on Windows.

tests/tools/console.test.ts
freema: Agreed, left them alone here so the diff stays focused. Happy to remove them in a follow-up.

juliandescottes (approved):
Thanks @freema 

Really nice to have Windows running on CI now.

PR #171, opened 2026-08-24
refactor: centralize tool handler error handling

f3tchcodes:
This change moves the repeated tool handler `try/catch` logic into a shared `defineToolHandler` wrapper. Errors are still returned using the same MCP error response format.

Catches with special behavior, such as UID errors, Firefox permission errors, and context cleanup, remain inside their handlers.

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -218,12 +218,12 @@ export const handleNewPage = defineToolHandler(async function handleNewPage(
     throw new Error('url parameter is required and must be a string');
   }
 
+  const waitFor = parseWait(wait);
+
   const { getFirefox } = await import('../index.js');
   const firefox = await getFirefox();
 
-  const newIdx = await firefox.createNewPage(url);
-
-  const waitFor = parseWait(wait);
+  const newIdx = await firefox.createNewPage(url, waitFor);
 
   return successResponse(`new page [${newIdx}] → ${url}${waitSuffix(waitFor)}`);
 });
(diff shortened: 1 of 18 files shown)


juliandescottes (changes requested):
Alright, although we have 1 new tool being added in PR #168, I think we can move forward. Can you rebase the PR here?

src/tools/pages.ts
```diff
@@ -196,152 +196,145 @@ function formatPageList(
...
+  const page = tabs[selectedIdx];
 
-    if (!page) {
-      throw new Error('No page selected');
-    }
+  const waitFor = parseWait(wait);
 
-    await firefox.navigate(url, waitFor);
+  // Refresh tabs to get latest list
+  await firefox.refreshTabs();
```
juliandescottes: `parseWait` needs to be before `const { getFirefox } = await import('../index.js');` and `await firefox.refreshTabs();` should only be called once.

That should be 
```ts
  const waitFor = parseWait(wait);

  const { getFirefox } = await import('../index.js');
  const firefox = await getFirefox();

  // Refresh tabs to get latest list
  await firefox.refreshTabs();
  const tabs = firefox.getTabs();
  const selectedIdx = firefox.getSelectedTabIdx();
  const page = tabs[selectedIdx];
```

(cannot post that as a code suggestion, because github is all confused about the diff apparently)

juliandescottes (changes requested):
Congrats on your first merge conflict resolution 🎉 . One small issue to address and this should be ok.

src/tools/pages.ts
f3tchcodes: This should be fixed now!

juliandescottes (approved):
Thanks for the update, looks good to me @f3tchcodes !

PR #172, opened 2026-08-25
fix(windows): handle file paths correctly on Windows

freema:
Split out of #169, which @juliandescottes asked me to break into smaller pieces. This part is independent of everything else in that PR.

Two fixes around file paths:

- The `saveTo` boundary check compared the resolved path against its allowed root with a case-sensitive `startsWith`. Windows paths are case-insensitive, so `c:\Users\me\.firefox-devtools-mcp\out.json` was rejected as "outside the allowed location" while the same path with a capital drive letter was accepted. Agents produce either spelling. Ignoring case on Windows only removes false rejections, it cannot let through a path that previously escaped.

- Both log paths opened their file without ensuring the directory existed, while the auto-generated path right beside one of them already called `mkdirSync`. An `--output-file` whose directory was missing threw ENOENT from deep inside `connect()`, and a `--log-file` in the same state silently disabled logging through a stream error. Not platform-specific, just easy to hit on Windows where `/tmp/foo.log` does not exist at all.

Verified on Windows 11, Node 22.22.0, Firefox 154.0. Every case was reproduced against `main` first:

| case | `main` | this branch |
| --- | --- | --- |
| `saveTo` with a lowercase drive letter (`c:\Users\...`) | rejected | accepted |
| `saveTo` with an uppercased segment (`\USERS\`) | rejected | accepted |
| `--log-file` into a missing directory | logging silently disabled | directory created, lines written |
| `--output-file` into a missing directory | ENOENT out of `connect()` | Firefox starts, 374 KB captured |

The escapes still fail closed: a path outside the root, the `~/.firefox-devtools-mcp-evil` prefix look-alike, a `..` traversal, and the same traversal spelled with a lowercase drive letter are all still rejected. Unit suite, typecheck, lint and `format:check` are green on Windows.

### DIFF
--- a/src/utils/save-output.ts
+++ b/src/utils/save-output.ts
@@ -30,6 +30,20 @@ async function isDirectory(path: string): Promise<boolean> {
 }
 
 /**
+ * Whether `candidate` is `root` or sits inside it. Compared case-insensitively
+ * on Windows, where `c:\Users\...` and `C:\Users\...` are the same directory —
+ * matching case exactly would reject valid paths, not catch escapes.
+ */
+export function isWithinRoot(root: string, candidate: string): boolean {
+  const normalize = (path: string) => (process.platform === 'win32' ? path.toLowerCase() : path);
+  const normalizedRoot = normalize(root);
+  const normalizedCandidate = normalize(candidate);
+  return (
+    normalizedCandidate === normalizedRoot || normalizedCandidate.startsWith(normalizedRoot + sep)
+  );
+}
+
+/**
  * Reject saveTo paths that escape the allowed roots, unless the server was
  * started with --unrestricted-save-paths. Relative paths must stay within the
  * current working directory; absolute paths must stay within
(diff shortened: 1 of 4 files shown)


src/utils/save-output.ts
```diff
@@ -35,13 +35,27 @@ async function isDirectory(path: string): Promise<boolean> {
  * current working directory; absolute paths must stay within
  * ~/.firefox-devtools-mcp.
  */
```
juliandescottes: This comment should stay close to `assertAllowedPath`

juliandescottes (changes requested):
Thanks, LGTM, one misplaced comment to fix.

juliandescottes (approved):
Thanks!

src/utils/save-output.ts
freema: Moved `isWithinRoot` above the docblock and squashed it into the first commit. Thanks.

PR #174, opened 2026-08-25
fix(windows): auto-detect Firefox binary for per-user installs

freema:
Split out of #169, which @juliandescottes asked me to break into smaller pieces. This is the part that needs a decision rather than just a review.

geckodriver only searches the Program Files directories and HKEY_LOCAL_MACHINE for Firefox, so it cannot see an install made without administrator rights: that one lands in `%LOCALAPPDATA%\Mozilla Firefox` and registers under HKCU. Launching then fails with "Expected browser binary location, but unable to find binary in default location" even though Firefox is installed and working.

You pointed at https://bugzilla.mozilla.org/show_bug.cgi?id=1921933 and you are right that this belongs in geckodriver. I am happy to take that mentored bug. The open question is whether you want an interim workaround in the MCP in the meantime, or would rather wait for the upstream fix and close this. I have referenced the bug in the module header either way, so it is clear this is meant to go away.

Changes from the review on #169:

- shortened the not-found message to your wording and moved the constant into `core.ts`, its only caller
- renamed `binaryLookupFailed` to `windowsBinaryLookupFailed`
- the module and its test now use `path.win32` instead of the ambient `path`. That was the failing check on #169: the suite mocks `process.platform` but ran on a Linux runner, so PATH was split on `:` instead of `;` and the lookup returned null.

Verified on Windows 11, Node 22.22.0, against a real per-user Firefox 154.0: nothing in either Program Files directory, nothing under HKLM, only `%LOCALAPPDATA%\Mozilla Firefox` and the HKCU registration. On `main` that machine fails with the "unable to find binary in default location" message above. On this branch:

- Firefox launches in 1.1 s and `get_firefox_info` reports `Binary: C:\Users\...\AppData\Local\Mozilla Firefox\firefox.exe (auto-detected)`
- the full server over stdio works end to end: 32 tools listed, `navigate_page` and `evaluate_script` both fine
- lookup order holds: with a Program Files install present it wins over the per-user one, matching geckodriver
- the fallbacks work in order, PATH then HKCU, and the registry lookup still resolves with an empty PATH, which is what invoking `reg.exe` by absolute path is for
- with nothing findable, the error names `--firefox-path` and keeps the original geckodriver message
- an explicit `--firefox-path` skips detection entirely

One gap worth knowing about: a PATH entry wrapped in quotes (`"C:\some dir";...`) is not unquoted before probing, so that entry is skipped. It falls through to the registry rather than failing, and I left it alone since this module is meant to be short-lived, but say the word and I will strip them.

### DIFF
--- a/src/firefox/core.ts
+++ b/src/firefox/core.ts
@@ -18,6 +18,7 @@ import { homedir } from 'node:os';
 import { dirname, join, delimiter } from 'node:path';
 import type { FirefoxLaunchOptions } from './types.js';
 import { log, logDebug } from '../utils/logger.js';
+import { findFirefoxBinaryWindows } from './windows-binary.js';
 import { resolveProfilePath } from './profile.js';
 
 // ---------------------------------------------------------------------------
@@ -150,8 +151,14 @@ async function findGeckodriver(): Promise<string> {
   return found;
 }
 
+/** geckodriver's "unable to find binary" error never mentions the fix. */
+const BINARY_NOT_FOUND_HINT =
+  'Unable to detect Firefox binary automatically, please provide the full path via ' +
+  '--firefox-path';
+
 export class FirefoxCore {
   private currentContextId: string | null = null;
+  private detectedBinaryPath: string | undefined;
   private driver: WebDriver | null = null;
   private firefoxVersion: string | null = null;
   private logFileFd: number | undefined;
@@ -319,6 +326,14 @@ export class FirefoxCore {
       }
       if (this.options.firefoxPath) {
         firefoxOptions.setBinary(this.options.firefoxPath);
+      } else if (process.platform === 'win32') {
+        // geckodriver misses per-user Windows installs.
+        const discoveredBinary = findFirefoxBinaryWindows();
+        if (discoveredBinary) {
+          logDebug(`Using auto-detected Firefox binary: ${discoveredBinary}`);
+          firefoxOptions.setBinary(discoveredBinary);
+          this.detectedBinaryPath = discoveredBinary;
+        }
       }
       if (this.options.args && this.options.args.length > 0) {
         firefoxOptions.addArguments(...this.options.args);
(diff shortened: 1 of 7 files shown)


src/firefox/core.ts
```diff
@@ -319,6 +329,16 @@ export class FirefoxCore {
       }
       if (this.options.firefoxPath) {
         firefoxOptions.setBinary(this.options.firefoxPath);
+      } else {
```
juliandescottes: Let's make that branch windows-only 
```suggestion
      } else if (process.platform === 'win32') {
```

And since binary.ts is a workaround while waiting for a geckodriver fix, let's also update it to clearly be only about windows. windows-binary.ts, and only expose `findFirefoxBinaryWindows`

src/firefox/core.ts
```diff
@@ -374,11 +394,23 @@ export class FirefoxCore {
...
+          .build();
+      } catch (error) {
+        if (windowsBinaryLookupFailed) {
+          // geckodriver's message never mentions --firefox-path.
+          throw new Error(
+            `${BINARY_NOT_FOUND_HINT}\n\nOriginal error: ${
+              error instanceof Error ? error.message : String(error)
+            }`
+          );
+        }
```
juliandescottes, line 400: Even with this flag, we can't be sure the error is related to the missing binary. 
geckodriver will look into locations that binary.ts doesn't check. 

We can imagine:
- binary.ts fails to find anything, `windowsBinaryLookupFailed = true`
- then geckodriver finds a binary
- `new Builder()...` fails for any other reason

-> We would still print an error message suggestion to use --firefox-path, which might be unhelpful. 

I would check if the geckodriver error message contains `moz:firefoxOptions.binary` (that's the capability name for the binary path, only used in that error message and very unlikely to change)

```
if (message.includes('moz:firefoxOptions.binary')) {
```

and I would drop `windowsBinaryLookupFailed` entirely. The missing-binary case isn't Windows-specific: on Linux geckodriver only searches PATH, so an install under /opt hits the same error. No reason to restrict the message to windows only.

juliandescottes (changes requested):
Thanks @freema , approach looks good, I think the new error message could also be helpful on other platforms. See suggestions in comments.

src/firefox/core.ts
freema: Done: renamed to windows-binary.ts, only findFirefoxBinaryWindows is exported and core.ts calls it behind a win32 check.

src/firefox/core.ts
freema: Done. The flag is gone; the hint now keys off moz:firefoxOptions.binary in the geckodriver message on every platform. Added unit tests for the hint and for other errors passing through unchanged.

juliandescottes (approved):
Thanks!

PR #175, opened 2026-08-27
chore: ignore typescript config files in eslint

f3tchcodes:
Ignoring vitest config in eslint to avoid parsing errors with the ESlint extension.

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -52,6 +52,10 @@
   },
   "overrides": [
     {
+      "files": ["*.config.ts", "*.config.*.ts"],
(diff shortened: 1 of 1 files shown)


juliandescottes (changes requested):
We have the same issue with all the config.ts files. I suggest to add something broader but which preserves linting. In the "overrides":

```
    {
      "files": ["*.config.ts", "*.config.*.ts"],
      "extends": ["plugin:@typescript-eslint/disable-type-checked"]
    }
```

juliandescottes (approved):
Great, thanks!

PR #176, opened 2026-08-27
chore: remove deprecated vitest poolOptions configurations

f3tchcodes:
Vitest 4 rewrote its pool system and removed `poolOptions` entirely. The repo upgraded to **Vitest 4.1.8** but its config wasn't fully migrated.

Causing this warning to show up when `npm run test:unit` is ran:
image

This change removes the deprecated vitest configurations from `vitest.config.ts`.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -11,11 +11,6 @@ export default defineConfig({
     // Run tests sequentially to avoid Firefox port conflicts
     fileParallelism: false,
     pool: 'forks',
-    poolOptions: {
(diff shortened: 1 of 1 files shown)


juliandescottes (approved):
Good catch! Looking at https://vitest.dev/guide/migration.html#pool-rework it seems that `singleFork` should instead be expressed via `maxWorkers: 1`. But apparently that's already enforced via `fileParallelism: false` so we can leave it as is!

PR #177, opened 2026-08-28
fix: correct tsup config types

f3tchcodes:
Fixing TypeScript errors in the tsup configuration caused by using `as const` on config values such as `format`.

image

The config objects now use `satisfies Options`, which keeps type inference while making sure the config uses valid tsup types.


juliandescottes (approved):
Thanks for the fix. We could probably add a CI check for those config files as well, but maybe not worth the effort / maintenance.

PR #178, opened 2026-08-28
feat: add explicit system access flag

f3tchcodes:
This change replaces hardcoded privelege access for public and local builds of Firefox with a new easier and straightforward to use `--allow-system-access` flag for explicitly enabling privileged Firefox access in the public build.

When enabled, the MCP allows privileged modules and launches Firefox with `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1`. Tools still need to be selected with `--tool-preset mozilla` or `--tools`.

Without the flag, privileged modules stay blocked and system access cannot be enabled through `--env` or `restart_firefox`.

**Some context:** I initially thought of adding this after Codex installed an extension using the public build, then spent around 10 minutes running PowerShell commands trying to find the extension UUID to access it's options page until all my tokens had been burnt out. I later realized `list_extensions` already does exactly this, but it was unavailable because privileged tools are completely blocked in the public build and Firefox also needs `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1` with the local build.

I am aware that system access is a powerful and potentially dangerous permission, but I think accessing privileged tools should at least be available as an explicit option rather than being completely blocked in the public build.

I have also went ahead and performed these tests to make sure everything works correctly and there are no major security concerns:-
- Privileged tools stay blocked without `--allow-system-access`.
- Privileged tools work with `--allow-system-access --tool-preset mozilla`.
- System access stays enabled after `restart_firefox` if it was launched with `--allow-system-access`.
- `restart_firefox` rejects `MOZ_REMOTE_ALLOW_SYSTEM_ACCESS=1` through `env`.

I believe this could serve as a temporary solution until WebDriver BiDi has a way to list installed extensions without requiring system access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Browser MCP servers carry inherent risks. A few key practices:
 
 - **Use a dedicated Firefox profile.** Never run the server against your regular profile — the agent has access to whatever the browser can reach, including cookies and saved sessions.
 - **Be cautious about which sites you visit.** Pages can return content designed to manipulate the agent (prompt injection). Stick to sites you control or trust.
-- **Enable only the tool modules you need.** The default `basic` preset already includes `evaluate_script`; `--tool-preset slim` drops it. Higher presets such as `--tool-preset developer` (debugging, network, console, profiler) and `--tool-preset mozilla` (privileged context) expand what the agent can do further.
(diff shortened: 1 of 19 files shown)


juliandescottes (changes requested):
Thanks for the detailed writeup and for listing out the tests you ran — and sorry about the token burn hunting for that UUID.

That said, I don't think we should change this boundary. All privileged code execution requiring the moz- package is intentional, and I'd rather keep it a hard line than make it opt-in behind a flag. What we should fix instead is making list_extensions available to the regular package. That means adding a new WebDriver BiDi command that doesn't require parent process access, which would go in:
https://searchfox.org/firefox-main/source/remote/webdriver-bidi/modules/root/webExtension.sys.mjs

I just filed https://bugzilla.mozilla.org/show_bug.cgi?id=2068045 to handle that. Are you interested in contributing this instead?

PR #179, opened 2026-08-30
feat: disconnect idle existing Firefox sessions

f3tchcodes:
Bugzilla: [Bug 2066912](https://bugzilla.mozilla.org/show_bug.cgi?id=2066912)

This change disconnects from an existing Firefox instance after 30 minutes of inactivity when using `--connect-existing`. The Firefox process itself stays open, and the MCP reconnects to the same instance on the next tool call. The idle timeout is reset by new tool activity and does not run while a tool call is still in progress.

I tested it manually by:
1. connecting to an existing Firefox and running a tool via inspector
2. waiting for the idle timeout and confirming geckodriver exits while Firefox stays open (you can refer to background processes for this)
3. calling another tool and confirming the MCP reconnects to the Firefox instance (it should reconnect to that same instance, but that specifically is not confirmed yet)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ import { parsePrefs, defaultProfileDir } from './cli.js';
 import type { parseArguments } from './cli.js';
 import { FirefoxDevTools } from './firefox/index.js';
 import type { FirefoxLaunchOptions } from './firefox/types.js';
+import { ExistingFirefoxIdleController } from './firefox/idle-connection.js';
 import { buildToolset } from './tools/registry.js';
 import { errorResponse } from './utils/response-helpers.js';
 
@@ -65,6 +66,10 @@ export function setNextLaunchOptions(options: FirefoxLaunchOptions): void {
  */
 export function isFirefoxRunning(): boolean {
   return firefox !== null;
+}
+
+function hasActiveExistingFirefoxConnection(): boolean {
+  return firefox?.getOptions().connectExisting === true;
 }
 
 /**
@@ -179,6 +184,13 @@ export async function run(
 
   args = parseArgsFn(SERVER_VERSION);
 
+  const idleConnection = new ExistingFirefoxIdleController({
+    enabled: Boolean(args.connectExisting),
+    hasActiveConnection: hasActiveExistingFirefoxConnection,
+    disconnect: resetFirefox,
+    onDisconnectError: (error) => logError('Error disconnecting idle Firefox session', error),
+  });
+
   if (args.logFile) {
     setupLogFile(args.logFile);
   }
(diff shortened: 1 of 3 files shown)


juliandescottes (changes requested):
Thanks, I would like to see a simpler implementation. The idle-connection wrapper is a bit too complex for something which only matters in one specific scenario.

I suggest to drop this wrapper and instead create a timer/promise/counter inline in `CallToolRequestSchema`. Downside is that we can't easily cover that with tests, but I think the simplification makes it worth this.

tests/firefox/connect-existing.test.ts
```diff
@@ -42,6 +42,72 @@ describe('getFirefox() reconnect behavior', () => {
     expect(core.getCurrentContextId()).toBe(null);
     expect(() => core.getDriver()).toThrow('Driver not connected');
   });
+
+  it('should keep retrying the pinned target after a reconnect failure', async () => {
```
juliandescottes: `existingFirefoxReconnectOptions` is never cleared when a reconnect fails so if I quit and reopen Firefox during the idle window, every following tool call keeps trying the dead port and will fail until I restart the MCP server. 

`restart_firefox` doesn't help either, it errors out before setNextLaunchOptions() when there's no --firefox-path.

I would clear `existingFirefoxReconnectOptions` on failure and let the next call go through lookupMarionettePort() again, it already picks the most recent live instance. 

This test will have to be updated accordingly.

src/firefox/core.ts
```diff
@@ -438,6 +405,14 @@ export class FirefoxCore {
     this.currentContextId = await this.driver.getWindowHandle();
     logDebug(`Browsing context ID: ${this.currentContextId}`);
 
+    if (resolvedMarionettePort !== undefined) {
+      this.options = {
+        ...this.options,
+        marionettePort: resolvedMarionettePort,
+        lookupMarionettePort: false,
```
juliandescottes: See L238, we use the fact that `lookupMarionettePort` is set detect if we are in the mode where the AI assistant companion is used in Firefox.

If we set this to false here, we will show an error message asking to pass a --marionette-port, whereas in this case most likely the user has to enable remote control again. 

 Can you keep the resolved port in a separate option and leave `lookupMarionettePort` as is?

src/index.ts
```diff
@@ -230,15 +238,22 @@ export async function run(
...
-  mcpServer.server.setRequestHandler(ListToolsRequestSchema, async () => {
+  server.setRequestHandler(ListToolsRequestSchema, async () => {
     log('Listing available tools');
     return {
       tools: allTools,
     };
   });
 
+  const connectExisting = Boolean(args.connectExisting);
+  const timeout = 30 * 60 * 1000; // 30 minutes
```
juliandescottes: The `firefox?.getOptions().connectExisting === true ` check below should already provide this information, you may drop the connectExisting flag.
```suggestion
  const idleTimeoutDuration = 30 * 60 * 1000; // 30 minutes
```

+ nit for the variable name

juliandescottes (changes requested):
Thanks @f3tchcodes some additional comments, looks close to landing.

juliandescottes (approved):
Looks good to me, thanks @f3tchcodes !

PR #186, opened 2026-09-04
chore: add webdriver-bidi-protocol types

hbenl:


src/firefox/events/network.ts
```diff
@@ -87,12 +88,13 @@ export class NetworkEvents {
...
-        method: req.request?.method || 'GET',
+        url: req.request.url,
+        method: req.request.method,
         timestamp: Date.now(),
-        resourceType: this.guessResourceType(req.request?.url || ''),
-        isXHR: req.initiator?.type === 'xmlhttprequest' || req.initiator?.type === 'fetch',
-        requestHeaders: this.parseHeaders(req.request?.headers || []),
+        resourceType: this.guessResourceType(req.request.url),
+        isXHR:
+          req.request.initiatorType === 'xmlhttprequest' || req.request.initiatorType === 'fetch',
```
hbenl, line 96: I switched from `req.initiator.type` to `req.request.initiatorType` because only that property has the values `xmlhttprequest` and `fetch` and `req.initiator.type` is [deprecated](https://www.w3.org/TR/webdriver-bidi/#cddl-type-networkinitiator).

tsconfig.json
```diff
@@ -28,7 +28,8 @@
     "allowUnreachableCode": false,
     "exactOptionalPropertyTypes": true,
     "noImplicitOverride": true,
-    "useUnknownInCatchVariables": true
+    "useUnknownInCatchVariables": true,
+    "isolatedModules": true
```
hbenl, line 32: We use vitest which uses esbuild and that always works as if `isolatedModules` was `true`. So I set it here to ensure that typescript warns us about any issues this may cause (like trying to access ambient const enums from `webdriver-bidi-protocol`).

src/firefox/index.ts
```diff
@@ -430,11 +435,11 @@ export class FirefoxClient {
    * @param behavior 'allowed' saves downloads silently, 'denied' cancels them, 'default' resets
    */
   async setDownloadBehavior(behavior: 'allowed' | 'denied' | 'default'): Promise<void> {
-    const downloadBehavior =
+    const downloadBehavior: Browser.DownloadBehavior | null =
       behavior === 'default'
         ? null
         : behavior === 'allowed'
-          ? { type: 'allowed' }
+          ? ({ type: 'allowed' } as unknown as Browser.DownloadBehavior)
```
hbenl, line 442: See [Bug 2069308](https://bugzilla.mozilla.org/show_bug.cgi?id=2069308)

src/firefox/events/network.ts
juliandescottes: ah good catch, thanks!

src/firefox/index.ts
juliandescottes: Nice find, the firefox implementation is wrong as well!

juliandescottes (approved):
Thanks Holger!

PR #187, opened 2026-09-06
chore: replacing deprecated server with McpServer

f3tchcodes:


src/index.ts
```diff
@@ -216,7 +216,7 @@ export async function run(
     logDebug(`  Viewport: ${args.viewport.width}x${args.viewport.height}`);
   }
 
-  const server = new Server(
+  const server = new McpServer(
```
juliandescottes: Maybe rename to `mcp` or `mcpServer` to avoid the `server.server` bits?

src/index.ts
```diff
@@ -230,15 +230,15 @@ export async function run(
   );
 
   // List available tools
-  server.setRequestHandler(ListToolsRequestSchema, async () => {
+  server.server.setRequestHandler(ListToolsRequestSchema, async () => {
```
juliandescottes: > @deprecated Use `McpServer` instead for the high-level API. Only use `Server` for advanced use cases.

I don't really understand the deprecation policy here. `Server` is "deprecated" unless you want to use advanced APIs. And apparently `setRequestHandler` is one of those? Is it then better to keep instantiating `Server` or to access it via `MCPServer.server`? Will one or both be removed in the future? 

Let's move forward with this for now, but it's not really clear to me what the guideline is.

(no action needed here, but let's be mindful of this if other "deprecation" cleanups come up in the future)

juliandescottes (changes requested):

src/index.ts
f3tchcodes: Yeah now that you mention it, that's confusing to me as well. My understanding is that `McpServer` is now the intended high level entry point while lower level methods like `setRequestHandler` are still exposed through `McpServer.server`. The deprecation wording is still a little uncertain about whether `Server` is completely deprecated or the access method is slightly changed (though probably the later).

src/index.ts
f3tchcodes: Done!

juliandescottes (approved):
Perfect, thanks!

PR #194, opened 2026-09-16
Bug 2072748 - chore: remove unused selector-based input methods

juliandescottes:


juliandescottes:
A bit of archeology here, the helpers were added speculatively in https://github.com/mozilla/firefox-devtools-mcp/commit/a2a9f2c but never actually used. A commit on the next day started implementing actions using snapshot instead of selectors.

hbenl (approved):

